PR #2, opened 2022-07-27
Add docstrings and comments

LuEdRaMo:


lbenet:
Hi @LuEdRaMo! Thanks for opening this PR to improve the docstrings/comments of the code, and sorry for the long delay to react... I was on holidays 😄. In the next few days I'll review this PR and send you some feedback. If you still are in the development process, let me know it, and then I'll review when you think it is ready.

src/asteroid_dynamical_models.jl
```diff
@@ -41,300 +67,462 @@ end
...
+    local newtonianNb_Potential_t = evaleph(params[3], dsj2k, q[1]) 
+    # Type of position / velocity components
     local S = eltype(q)
-    local UJ_interaction = params[5] # interaction matrix with flattened bodies
-    local N = params[6] # number of bodies, including NEA
+    # Interaction matrix with flattened bodies
+    local UJ_interaction = params[5] 
+    # Number of bodies, including NEA
+    local N = params[6] 
+    # Number of bodies, except Apophis
```
lbenet, line 119: Instead of "Apophis" I think it is better to write "asteroid" or "NEO"; the first suggestion is used in other files. Note that the same comment applies below and to other files.

src/process_radar_data_jpl.jl
```diff
@@ -62,65 +84,111 @@ rcvr(rdata::RadarDataJPL) = rdata.rcvr
 xmit(rdata::RadarDataJPL) = rdata.xmit
 bouncepoint(rdata::RadarDataJPL) = rdata.bouncepoint
 
-#monostatic mode: check that each receiver and transmitter are the same
+@doc raw"""
+    ismonostatic(rdata::RadarDataJPL)
+
+Checks wheter the setup is monostatic, i.e., reciever and transmitter are the same.
```
lbenet, line 90: wheter -> whether
reciever -> receiver

src/propagation.jl
```diff
@@ -1,143 +1,273 @@
...
-        acc_eph.x[:,3j-2] .= differentiate.(ss16asteph.x[:,3(Nm1+j)-2])
-        acc_eph.x[:,3j-1] .= differentiate.(ss16asteph.x[:,3(Nm1+j)-1])
-        acc_eph.x[:,3j  ] .= differentiate.(ss16asteph.x[:,3(Nm1+j)  ])
-    end #for, j
+            end
+        end
+        # Fill acelerations by differentiating velocities 
+        acc_eph.x[:,3j-2] .= differentiate.(ss16asteph.x[:,3(Nm1+j)-2])  # X-axis component
+        acc_eph.x[:,3j-1] .= differentiate.(ss16asteph.x[:,3(Nm1+j)-1])  # Y-axis component
+        acc_eph.x[:,3j  ] .= differentiate.(ss16asteph.x[:,3(Nm1+j)  ])  # Z-axis component
```
lbenet, line 74: Should `differentiate` be used here, or the specialized method defined in PlanetaryEphemeris.jl? (cc: @PerezHz )

lbenet (commented):
First, my apologies for the long time to review this PR... This PR is somewhat close to https://github.com/PerezHz/PlanetaryEphemeris.jl/pull/2 and thus some suggestions/comments apply here as well. This is part of the reason I only send few comments in the code.

The following are my general comments/suggestions:
- Move all `const`s to a `constants.jl` file: relevant for 'src/NEOs.jl',  'src/asteroid_dynamical_models.jl', and 'src/initial_conditions.jl'
- Write "asteroid" instead of "Apophis"; I think it is more general. This does not apply to src/delay_doppler.jl, since the files used there are specific to Apophis. TODO: Think on a "more general" interface, not specific for Apophis.
- Some functions in 'src/integration_methods.jl' include the name "apophis"; I think it is better to name them as "neo" or "asteroid", which matter for some notebooks and for the scripts. Perhaps we should open an issue with this proposal to make this change in another PR.

cc @PerezHz

src/process_radar_data_jpl.jl
LuEdRaMo: Done.

LuEdRaMo:
- I've moved most constants to a new file `src/constants.jl`.
- I've changed **apophis** for **asteroid** mostly in `src/asteroid_dynamical_models.jl`. In my opinion all other occurrences are apophis-specific. 
- Right now NEOs can't be compiled, since it requires v0.10 of [TaylorSeries](https://github.com/JuliaDiff/TaylorSeries.jl) while [PlanetaryEphemeris.jl](https://github.com/PerezHz/PlanetaryEphemeris.jl) requires v0.11. Therefore in the last commit I suggest to change [TaylorSeries](https://github.com/JuliaDiff/TaylorSeries.jl) compatibility to include v0.11.

Project.toml
```diff
@@ -40,7 +40,7 @@ RemoteFiles = "0.4"
 SPICE = "0.2"
 SatelliteToolbox = "0.9"
 TaylorIntegration = "0.8"
-TaylorSeries = "0.10"
+TaylorSeries = "0.10, 0.11"
```
lbenet, line 43: What about of extending this to include TaylorSeries v0.12, which is the most recent one?

lbenet (approved):
I have left a single comment, asking whether we should consider increasing the possible versions of TaylorSeries to include v0.12 or not.

A last question: did you check that everything behaves identically *after* moving all the constants to a file? 

Aside from that, LGTM and I am in favor of merging... but I wait for your answer to the question above before doing that. Thanks a lot again!

src/propagation.jl
PerezHz: Sorry for the late reply! I recall correctly, I avoided bumping TaylorSeries to the latest version in `NEOs.jl` precisely due to this issue, so probably it makes sense to use the other method (i.e., the `order`-preserving derivative), but we would have to check

LuEdRaMo:
- I've run `pha/apophis.jl` and the results of the integration do not change with the commits in this PR.
- I've executed `add PlanetaryEphemeris` in `Pkg` to update `Manifest.toml`. Before this change, `NEOs` could not be compilated.

Project.toml
LuEdRaMo: Currently, we cannot update `TaylorSeries` to the latest version v0.12 since `PlanetaryEphemeris` requires at most v0.11. Therefore, we need to update first in `PlanetaryEphemeris` and then in `NEOs`.

src/propagation.jl
LuEdRaMo: I've run `pha/apophis.jl` and the results do not change by using `PlanetaryEphemeris.ordpres_differentiate`.

LuEdRaMo:
I've run `pha/apophis.jl` again and the results do not change by using `PlanetaryEphemeris.ordpres_differentiate`.

src/asteroid_dynamical_models.jl
LuEdRaMo: I've changed apophis for asteroid mostly in `src/asteroid_dynamical_models.jl`. In my opinion all other occurrences are apophis-specific.

src/propagation.jl
lbenet: Excellent!

Project.toml
lbenet: TaylorSeries v0.12.2 includes a fix on `promotion_rule` that solves some ambiguities, and a bug in `tan`; the latter is more important in the context of PlanetaryEphemeris. I did a quick search for `tan` and I think it only appears in the computation of the osculating terms. If this is correct, there shouldn't be any problem to bump the version of TaylorSeries used in PlanetaryEphemeris and here, and we should perhaps have it 0.12 only.

lbenet:
> I've run pha/apophis.jl again and the results do not change by using PlanetaryEphemeris.ordpres_differentiate.

Excellent! I think we should decide about the minim version of TaylorSeries for both PlanetaryEphemeris and NEOs, and then merge the PRs, and bump to a *minor* version.

LuEdRaMo:
After the last commit, neither
- updating `TaylorSeries` to v0.12,
- updating `PlanetaryEphemeris` to v0.2,
- using order-preserving differentiate

changes the results of `pha/apophis.jl`.

lbenet (approved):
Thanks a lot, @LuEdRaMo! LGTM!

I think we should bump a new (minor) version, and then merge!

LuEdRaMo:
> Thanks a lot, @LuEdRaMo! LGTM!
> 
> I think we should bump a new (minor) version, and then merge!

Done.

lbenet:
Thanks a lot once again @LuEdRaMo for all the work !

Issue #3, opened 2022-11-01
Towards registering NEOs.jl


PerezHz:
While trying to register NEOs.jl in the Julia General registry (see relevant NEOs.jl [commit](https://github.com/PerezHz/NEOs.jl/commit/d64a74741983bdb0692dc14ad3e1a4166e711a89#commitcomment-88570260), corresponding PR JuliaRegistries/General/pull/71434 and [job log](https://github.com/JuliaRegistries/General/actions/runs/3372030329/jobs/5594980311)) auto-merge didn't work due to several issues:

- [x] Name of package is not at least 5 characters long
- [x] The following dependencies do not have a [compat] entry that is upper-bounded and only includes a finite number of breaking releases: DataFrames, HORIZONS, IntervalArithmetic, PlanetaryEphemeris, Query, ReferenceFrameRotations, StaticArrays, StatsBase
- [x] I was not able to install the package (i.e. Pkg.add("NEOs") failed). See the CI logs for details.
- [x] I was not able to load the package (i.e. import NEOs failed). See the CI logs for details.
- [ ] Package name similar to 2 existing packages.
  - Similar to NES. Damerau-Levenshtein distance 2 is at or below cutoff of 2. Damerau-Levenshtein distance 1 between lowercased names is at or below cutoff of 1.
  - Similar to NExOS. Damerau-Levenshtein distance 2 is at or below cutoff of 2. Damerau-Levenshtein distance 1 between lowercased names is at or below cutoff of 1.

I'm opening this issue to try to resolve those issues, so that we can finally register NEOs.jl.

Besides discussing the name of the package (I would be in favor of keeping NEOs, but maybe NearEarthAsteroids could be an alternative), I think the other important points would be to fix the compat entries and using consistent bounds on TaylorSeries.jl and TaylorIntegration.jl throughout so that `Pkg.add` and `Pkg.import` work.

cc: @lbenet @LuEdRaMo

PerezHz:
Actually, now that I think about it, since TaylorIntegration has a bound on TaylorSeries while re-exporting it, maybe we can get rid of the explicit dependency on TaylorSeries both here and in PlanetaryEphemeris.jl?

lbenet:
Thanks for opening this issue!

TLDR: upgrading TaylorSeries is ok; let's not do it yet for TaylorIntegration.

[...]

PR #4, opened 2022-11-03
Fix issues with registration

PerezHz:
This PR attempts to address some parts of #3, i.e., solve remaining issues blocking the registration of NEOs.jl, except the naming of the package itself. I worked around the `addrequire`s, which was giving some headaches; it works for now, but maybe later we can migrate to JLD2, which after @LuEdRaMo's investigations looks like it could be a promising alternative.

I propose to merge this and attempt to register this version; I would then expect auto-merge to be blocked only by the name of the package. If that happens, maybe we can appeal to a human about why the current name of the package makes sense; if the appealing fails then we can decide on a new name. I see there's #2 waiting to be merged; is there anything left to be done there, or should we go ahead and merge #2 before this one?

cc: @lbenet @LuEdRaMo


lbenet (commented):
I think we need to include in the `[compat]` section of the `Project.toml` file the dependency (with version) of `PlanetaryEphemeris`. It is the second item in the list in #3... 

Incidentally, I agree that it is better to simply delete the `Manifest.toml` file.

Aside from this, LGTM!

PerezHz:
> I think we need to include in the [compat] section of the Project.toml file the dependency (with version) of PlanetaryEphemeris. It is the second item in the list in #3...

Thank you for the heads up! After rebasing to current main it wasn't necessary to add it, since it was already added in #2 😄 

Alright, I'll then go ahead merging this PR and retry registering the package...

PR #13, opened 2023-06-02
Custom serialization for TaylorNs

LuEdRaMo:
This PR introduces a more efficient way to save the most common output produced by NEOs to `.jld2` files. As an example, a 100 years jet transport integration (+roots) of asteroid 2023 DW weights around 3.90 GB with master and 3.1 GB with this PR.


LuEdRaMo:
Thank you @PerezHz, I only have to implement a subtlety about the function that loads TaylorNs and this should be ready for review.

PerezHz:
Well, that's odd... Do the commits look as you expect in your local repo? Are you sure you pushed to the right remote with origin properly set and to the right branch?

LuEdRaMo:
In my local repo everything looks fine. Actually, if I check the `lerm/Custom_serialization` branch in this repo, I can see the latest commit `Corrections to TaylorN Serialization`.

LuEdRaMo:
I pushed a new commit and now the previous one (`Corrections to TaylorN serialization`) appears in this PR but not the new one.

lbenet:
The last commit I can see is "Bump minor version", which i think is also the last one in your branch.

LuEdRaMo:
Yes, It suddently corrected, I don't know if @PerezHz did some backstage magic...

PerezHz:
> Yes, It suddently corrected, I don't know if @PerezHz did some backstage magic...

No backstage magic I swear! 😅 Probably just a glitch in the (GitHub) matrix... 🐈‍⬛

LuEdRaMo:
I think this is ready, although I'll wait in case you have any comments.

PerezHz:
That's great! I'll try to review the code asap, but right now I have two questions: how large were the files without the TaylorN serialization? And how long are the 100 years integrations taking?

LuEdRaMo:
- Without this PR, a 100 years, 5-order, 6 variables jet transport integration is  3.90 GB. With this PR, the same file is 3.10 GB. So, its around a 20% gain. 
- The same integration takes around 9 hours in 10 threads.

src/propagation/serialization.jl
```diff
@@ -0,0 +1,204 @@
+@doc raw"""
+    TaylorInterpolantNSerialization{T}
+
+Custom serialization struct to save a `TaylorInterpolant{T, TaylorN{T}, 2}` to a `.jld2` file. 
+
+# Fields
+- `vars::Vector{String}`: jet transport variables. 
+- `order::Int`: order of Taylor polynomials.
```
PerezHz, line 8: I'd add here "w.r.t. time" at the end of the phrase

src/propagation/serialization.jl
```diff
@@ -0,0 +1,204 @@
...
+# Convert method to write .jld2 files
+function convert(::Type{TaylorInterpolantNSerialization{T}}, eph::TaylorInterpolant{T, TaylorN{T}, 2}) where {T}
+    # Variables 
+    vars = TS.get_variable_names()
+    # Number of variables 
+    n = length(vars)
+    # Matrix dimensions
+    dims = size(eph.x)
+    # Number of elements in matrix
+    N = dims[1] * dims[2]
```
PerezHz, line 37: Would this be equivalent to `N = length(eph.h)`? Most of the time I'd imagine both work out fine, but `length`, if it works, is maybe a bit more generic.

src/propagation/serialization.jl
```diff
@@ -0,0 +1,204 @@
...
+# Convert method to read .jld2 files
+function convert(::Type{TaylorInterpolant{T, TaylorN{T}, 2}}, eph::TaylorInterpolantNSerialization{T}) where {T} 
+    # Variables 
+    vars = eph.vars
+    # Number of variables 
+    n = length(vars)
+    # Matrix dimensions
+    dims = eph.dims
+    # Number of elements in matrix
+    N = dims[1] * dims[2]
```
PerezHz, line 82: Same as above.

test/propagation.jl
```diff
@@ -225,4 +246,54 @@ using InteractiveUtils: methodswith
         @test nrms(res, w) ≈ 0.375 atol=1e-2
     end
 
+    @testset "Jet transport propagation and TaylorN serialization" begin
+        # Test integration (Apophis)
+
+        # Dynamical function
+        local dynamics = RNp1BP_pN_A_J23E_J2S_eph_threads!
```
PerezHz, line 253: Minor aesthetic detail: I think a space is missing here at the beginning

PerezHz (approved):

PerezHz:
Many thanks @LuEdRaMo for adding this new feature together with tests! I left some comments here and there but overall LGTM! This `TaylorInterpolant` is kind of growing into a library by itself, probably it can make sense to put it into a separate package eventually, but I guess that has a lower priority right now.

src/propagation/serialization.jl
LuEdRaMo: Done.

src/propagation/serialization.jl
LuEdRaMo: I've changed it to `N = length(eph.x)`.

src/propagation/serialization.jl
LuEdRaMo: In this case, since we are reconstructing the matrix, there is no `x` to do `length(x)`.

test/propagation.jl
LuEdRaMo: Done.

test/propagation.jl
```diff
@@ -225,4 +246,54 @@ using InteractiveUtils: methodswith
...
+        recovered_tvS = JLD2.load("test.jld2", "tvS")
+        recovered_xvS = JLD2.load("test.jld2", "xvS")
+        recovered_gvS = JLD2.load("test.jld2", "gvS")
+        @test sol == recovered_sol
+        @test tvS == recovered_tvS
+        @test xvS == recovered_xvS
+        @test gvS == recovered_gvS
+        rm("test.jld2")
+
+        # It is unlikely that such a short integration generates a non-trivial tvS, xvS and gvS. Therefore, to test
```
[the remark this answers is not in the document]
LuEdRaMo: I did a few integrations and I think we can afford setting the initial condition close to Apophis close approach at April 13th 2029. In such setting, 10 steps are enough to catch a non trivial `gvS`.

LuEdRaMo:
I've addresed all the reviews and the tests pass. I'll go ahead and merge...

Issue #9, opened 2023-05-19
Add querying support for Vector{RadecMPC} and Vector{RadarJPL}


PerezHz:
Related to discussion around [this comment](https://github.com/PerezHz/NEOs.jl/pull/7#discussion_r1199289607)

PerezHz:
Fixed by #15

PR #15, opened 2023-06-10
DataFrames extension

LuEdRaMo:
In order to address https://github.com/PerezHz/NEOs.jl/issues/9, this PR changes `DataFrames` to a weakdep and introduces an  extension `DataFramesExt` with methods to convert between `Vector{AbstractAstrometry}` and `DataFrame`.


LuEdRaMo:
@PerezHz and @lbenet, since currently we only test on 1.9, should I use `Requires` to make the extension work with previous versions of julia?

PerezHz:
Many thanks for adding this, @LuEdRaMo! Is this ready for review, or are other things remaining to be included in this PR?

LuEdRaMo:
> Many thanks for adding this, @LuEdRaMo! Is this ready for review, or are other things remaining to be included in this PR?

I think it is ready for review.

ext/DataFramesExt.jl
```diff
@@ -0,0 +1,47 @@
...
+if isdefined(Base, :get_extension)
+    using DataFrames
+    import DataFrames: DataFrame
+else
+    using ..DataFrames
+    import ..DataFrames: DataFrame
+end
+
+# Method to convert a vector of observations to DataFrame
+function DataFrame(obs::Vector{T}) where {T <: AbstractAstrometry}
```
PerezHz, line 17: While this should work fine for most purposes, I wonder how it compares in terms of performance and memory use to other solutions via Tables.jl overloadings (e.g. [here](https://discourse.julialang.org/t/mapping-vector-mytype-to-a-dataframe/24797/13)) or via StructArrays.jl (e.g. [here](https://discourse.julialang.org/t/mapping-vector-mytype-to-a-dataframe/24797/14))? Doing some quick benchmarks could really help us to select between alternatives

PerezHz: (If indeed they apply, which they might not..)

ext/DataFramesExt.jl
```diff
@@ -0,0 +1,47 @@
...
+        push!(df, [getfield(obs[i], colnames[j]) for j in eachindex(colnames)])
+    end 
+
+    return df 
+end 
+
+# Methods to convert a DataFrame to a vector of AbstractAstrometry
+for T in (:(RadecMPC), :(RadarJPL))
+    @eval begin
+        function $T(df::DataFrame)
```
PerezHz, line 31: This comment is maybe just aesthetics, but while it makes sense to have this defined as some kind of external constructor as you have done, it feels like we should be able to do this via type parameters 🤔... maybe another option would be to define something like

```julia
    function dataframe(x::T) where {T<:AbstractAstrometry}
# ...
    end
```
?

test/extensions.jl
```diff
@@ -0,0 +1,32 @@
...
+    radec_2023DW = read_radec_mpc(joinpath("data", "RADEC_2023_DW.dat"))
+
+    df_2023DW = DataFrame(radec_2023DW)
+
+    @test nrow(df_2023DW) == length(radec_2023DW)
+    @test all(names(df_2023DW) .== String.(fieldnames(RadecMPC)))
+
+    recovered_2023DW = RadecMPC(df_2023DW)
+
+    @test recovered_2023DW == radec_2023DW
```
PerezHz, line 19: I would suggest to add a very simple test of selecting observations (one test for optical and one for radar) by observatory and checking that these kind of operations return what we expect. For example, let's say that we have 10 observations from observatory A on a given night, then we should be able to get these observations via a select on the DataFrame.

PerezHz (approved):
Again, many thanks for adding this! Great weakdeps use case and application! I added some comments regarding the conversions to and from DataFrames, but if there's no clear advantage of other approaches, then there's no need to change them. Also I think we could add simple select tests on the DataFrames just to make sure we have the essential functionality that we want. Overall, LGTM!

ext/DataFramesExt.jl
LuEdRaMo: I did the following benchmarks by converting `data/99942.dat` to `DataFrame`:
- It turns out that `DataFrames` already has a method that allows to convert a vector of struct to `DataFrame` and it performs as follows:
```julia
298.700 μs (323 allocations: 1.15 MiB)
```
- The method proposed by this PR is not quite efficient:
```julia
26.391 ms (384091 allocations: 25.54 MiB)
```
- By overloading `Tables.jl` we get and amazing performance:
```julia
193.000 μs (76 allocations: 1.14 MiB)
```
Althouh this would imply that the weakdep is `Tables.jl` instead of `DataFrames.jl`.
- By using `StructArrays.jl` we get a slightly worse performance:
```julia
324.000 μs (104 allocations: 2.27 MiB)
```

It is clear that overloading `Tables.jl` is the most efficient option but it requires changing the weakdep. @PerezHz , @lbenet , should I proceed with this change?

PerezHz: > It is clear that overloading Tables.jl is the most efficient option but it requires changing the weakdep. @PerezHz , @lbenet , should I proceed with this change?

Thanks for looking into this! I would say yes, let's go for it!

test/extensions.jl
LuEdRaMo: I've added two simple tests using `Query.jl` to filter observations by date and observatory.

ext/DataFramesExt.jl
```diff
@@ -0,0 +1,45 @@
...
+istable(::Type{Vector{RadecMPC{T}}}) where {T <: AbstractFloat} = true
+rowaccess(::Type{Vector{RadecMPC{T}}}) where {T <: AbstractFloat} = true
+rows(x::Vector{RadecMPC{T}}) where {T <: AbstractFloat} = x
+schema(::Vector{RadecMPC{T}}) where {T <: AbstractFloat} = Schema(fieldnames(RadecMPC{T}), Tuple{fieldtypes(RadecMPC{T})...})
+
+# Methods to convert a Vector{RadarJPL{T}} to a DataFrame
+istable(::Type{Vector{RadarJPL{T}}}) where {T <: AbstractFloat} = true
+rowaccess(::Type{Vector{RadarJPL{T}}}) where {T <: AbstractFloat} = true
+rows(x::Vector{RadarJPL{T}}) where {T <: AbstractFloat} = x
+schema(::Vector{RadarJPL{T}}) where {T <: AbstractFloat} = Schema(fieldnames(RadarJPL{T}), Tuple{fieldtypes(RadarJPL{T})...})
```
PerezHz, line 24: It's great that the Tables.jl interface works so well, many thanks for taking care of this! I think we can go a step beyond and further simplify this to something like:
```suggestion
# Methods to convert a Vector{T} where {T <: AbstractAstrometry} to a DataFrame
istable(::Type{Vector{<: AbstractAstrometry}}) = true
rowaccess(::Type{Vector{<: AbstractAstrometry}}) = true
rows(x::Vector{<: AbstractAstrometry}) = x
schema(::Vector{T}) where {T <: AbstractAstrometry} = Schema(fieldnames(T), Tuple{fieldtypes(T)...})
```

ext/DataFramesExt.jl
PerezHz: Actually, after taking a closer look, it appears as if `RadecMPC` and `RadarJPL` constructors are being defined here (i.e., methods with the name of a type), which then should return an instance of the type they're constructing, but these methods are actually returning `Vector{RadecMPC}` and `Vector{RadarJPL}`, respectively. I think this can be a bit confusing, so I wonder whether we could just define these constructors over a single row of a `DataFrame`, and handle the conversion of the full DataFrame via broadcasting? I mean something like

```julia
   RadecMPC(row_radec) = # constructor a RadecMPC from a single row of a DataFrame
   RadarJPL(row_radar) = # same for RadarJPL
   
   # then, to convert from DataFrame to RadecMPC, RadarJPL:
   RadecMPC.(df_radec)
   RadarJPL.(df_radar)
```

PerezHz: (Resolving this conversation in favor of a more concrete suggestion...)

ext/DataFramesExt.jl
```diff
@@ -0,0 +1,45 @@
...
+            @assert all(numtypes .== numtypes[1])
+            S = numtypes[1]
+            obs = Vector{$T{S}}(undef, nrow(df)) 
+            for i in eachindex(obs)
+                obs[i] = $T(values(df[i, :])...)
+            end 
+
+            return obs 
+        end 
+    end
```
PerezHz, line 42:
```suggestion
# Methods to convert a DataFrame to a Vector{RadecMPC{T}} / Vector{RadarJPL{T}}
function Vector{T}(df::DataFrame) where {T <: AbstractAstrometry}
    @assert all(string.(fieldnames(T)) .== names(df))
    coltypes = eltype.(eachcol(df))
    numtypes = filter!(x -> x <: AbstractFloat, coltypes)
    @assert all(numtypes .== numtypes[1])
    obs = Vector{T}(undef, nrow(df))
    for (i, row) in zip(eachindex(obs), eachrow(df))
        obs[i] = T(values(row)...)
    end

    return obs
end

convert(::Type{Vector{T}}, df::DataFrame) where {T <: AbstractAstrometry} = Vector{T}(df::DataFrame)
```

PerezHz: After thinking more about it, I would suggest (see suggestion) to define a `Vector{T} where T<:AbstractAstrometry` constructor, which I think it's conceptually clearer and avoids having duplicate methods which do essentially the same for both types of astrometry. I also suggest to add a `convert` method, which would allow julia to find the appropriate constructor, though I'm not 100% sure this last thing is needed. If we go for this route we would have to import Base.convert in the extension, and update the tests so that we construct from a DataFrame as `Vector{RadecMPC}(df)`.

PerezHz:
We've got the green lights, so we're GO for merging. Thanks a lot, @LuEdRaMo!

PR #16, opened 2023-06-14
Lower `propagate` and `compute_delay` latency

PerezHz:
This PR lowers the latency of `propagate` and `compute_delay`; still WIP.


PerezHz:
I've added some tests to check that radar astrometry computations work fine with jet transport. Otherwise, I'll go ahead and merge this (and bump the patch version).

PR #18, opened 2023-06-20
Fix backward propagation

LuEdRaMo:
Currently, backward propagation throws an error because this line:
https://github.com/PerezHz/NEOs.jl/blob/905a80e0c40640db0df0904726296d70c1ec2c0d/src/propagation/propagation.jl#LL324C1-L324C60
loads an incomplete time interval. This PR introduces a simple change to fix this.

### DIFF
--- a/src/observations/jpl_eph.jl
+++ b/src/observations/jpl_eph.jl
@@ -119,25 +119,24 @@ dtt_tdb(et) = getposvel(1000000001, 1000000000, cte(et))[4] # units: seconds/sec
 
 # Load Solar System 2000-2100 ephemeris
 const sseph_artifact_path = joinpath(artifact"sseph_p100", "sseph343ast016_p100y_et.jld2")
-const sseph = JLD2.load(sseph_artifact_path, "ss16ast_eph")
-const ttmtdb = TaylorInterpolant(sseph.t0, sseph.t, sseph.x[:,end])
+const sseph::TaylorInterpolant{Float64, Float64, 2} = JLD2.load(sseph_artifact_path, "ss16ast_eph")
+const ttmtdb::TaylorInterpolant{Float64, Float64, 1} = TaylorInterpolant(sseph.t0, sseph.t, sseph.x[:,end])
 
 @doc raw"""
-    loadpeeph(et::Union{Nothing, Real} = nothing)
+    loadpeeph()
+    loadpeeph(et::Real)
     loadpeeph(et_0::Real, et_f::Real)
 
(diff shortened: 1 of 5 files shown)


src/observations/jpl_eph.jl
```diff
@@ -124,20 +124,28 @@ const ttmtdb = TaylorInterpolant(sseph.t0, sseph.t, sseph.x[:,end])
...
 function loadpeeph(et::Union{Nothing, Real} = nothing)
     if isnothing(et)
         return sseph
     else
-        idxs = findall(x -> x <= et, sseph.t)
-        return TaylorInterpolant(sseph.t0, sseph.t[idxs], sseph.x[idxs[1:end-1], :])
+        i = searchsortedfirst(sseph.t, et)
+        return TaylorInterpolant(sseph.t0, sseph.t[1:i], sseph.x[1:i-1, :])
     end
 end
```
lbenet, line 141: Two comments (about this function):
- I think it makes sense to exploit dispatch here, and define the functions
    ```julia
    loadpeeph() = sseph
    loadpeeph(::Nothing) = loadpeeph()
    ```
    In this way, you avoid the the `Union`, the `if` and thus all branching.
- What is `sseph` here and in my previous suggestion? My point is that it is either a `global` quantity, which we should perhaps avoid, or simply undefined...

PerezHz: > I think it makes sense to exploit dispatch here, and define the functions

+ 1 to the suggestion of using multiple dispatch 👍

PerezHz:
> Currently, backward propagation throws an error because this line:

Woops, sorry about that! I suggest to add also backward propagation to the tests!

PerezHz:
And many thanks for the fix @LuEdRaMo!

src/observations/jpl_eph.jl
PerezHz: Regarding `sseph`, I suggest to type-annotate it, to help the compiler with type-inference:
```julia
const sseph::TaylorInterpolant{Float64,Float64,2} = JLD2.load(sseph_artifact_path, "ss16ast_eph")
```
We could even consider adding type annotations to `loadpeeph`:
```julia
loadpeeph()::TaylorInterpolant{Float64,Float64,2} = sseph
loadpeeph(::Nothing)::TaylorInterpolant{Float64,Float64,2} = loadpeeph()
```
The rationale behind using `sseph` is to load the the Solar System ephemeris  via artifacts only once in memory.

PerezHz:
I went ahead and included a couple of unit tests for backward propagations, and bumped the patch version. I propose to merge this and address further issues (e.g., #19) in separate PRs. Again, great catch and many thanks for this fix, @LuEdRaMo!

PR #24, opened 2023-06-25
Add support for multi-threaded reduction of observations (optical and radar)

PerezHz:
Now that we have our own TT-TDB ephemeris, this is now within reach of hand. The trick is to avoid calls to `SPICE.deltet` and `SPICE.str2et` and adding `Threads.@threads` to the for loop over the astrometry in `compute_delay` and `compute_radec`. I'm still testing some stuff, but this should be ready for review soon.


PerezHz:
Tests are passing, so I'll go ahead and merge this so that we can tackle the update to PlanetaryEphemeris v0.7.

Issue #17, opened 2023-06-18
Precompute accelerations and Newtonian potentials


PerezHz:
This issue follows from an offline discussion with @lbenet and @LuEdRaMo. Currently, before performing a propagation, we use `loadeph` to precompute the accelerations and the Newtonian potentials of the perturbers (planets, asteroid, etc.) of a NEO orbit, besides changing the units of the ephemeris themselves from km and km/sec to au, au/day. But now we have `NEOs.sseph`, which would allow us to precompute these things, so that their computation can be taken out from the integration loop. This can help simplify code (and maybe improve performance?).

LuEdRaMo:
As a reference, the files we need to have precomputed accelerations and Newtonian potentials as artifacts are: 
- 100 years Solar System integration (556 MB).
- corresponding accelerations (257 MB).
- corresponding Newtonian potentials (86 MB)

Together, they fit within the 1GB of free Github LFS.

LuEdRaMo:
Currently, `NEOs.sseph` has time in seconds but we convert to days for the integration. Should we skip this conversion by saving `NEOs.sseph` in days in the first place?

PerezHz:
> Currently, NEOs.sseph has time in seconds but we convert to days for the integration. Should we skip this conversion by saving NEOs.sseph in days in the first place?

Thank you @LuEdRaMo for checking this, it's great news that we can fit everything within the available GitHub LFS space 😅! And yeah, I agree we should save `NEOs.sseph` in au, au/day, although then we have to be careful when computing observables, since we are now assuming that they read ET seconds since J2000 and ouput pos. and vel., resp., in km and km/sec.

PerezHz:
I would propose to first push #15 and #16 over the finish line and then tackle this.

PerezHz:
Fixed by #25.

PR #25, opened 2023-06-28
Update to PlanetaryEphemeris v0.7

PerezHz:
I’m opening this PR to join efforts to update to PE v0.7. We should be now in a position to update the ephemeris in LuEdRaMo/sseph and proceed to the corresponding changes here.

cc: @lbenet @LuEdRaMo


LuEdRaMo:
`LuEdRaMo/sseph` has been updated to PE v0.7 ephemeris, so we have green flag to start this PR.

LuEdRaMo:
With the last commit, propagation methods now work with PE v0.7 and with precomputed accelerations and newtonian potentials, so this PR should solve https://github.com/PerezHz/NEOs.jl/issues/17. Also, `propagate` and `propagate_root` now use `taylorinteg` instead of `neosinteg`. Currently, `propagate_root` cannot undergo the same change since root-finding `taylorinteg` has no dense output option.

lbenet:
There are some changes coming in TaylorIntegration; see https://github.com/PerezHz/TaylorIntegration.jl/pull/168. Could you have a look if everything works smoothly?

PerezHz:
> With the last commit, propagation methods now work with PE v0.7 and with precomputed accelerations and newtonian potentials, so this PR should solve https://github.com/PerezHz/NEOs.jl/issues/17. Also, propagate and propagate_root now use taylorinteg instead of neosinteg. Currently, propagate_root cannot undergo the same change since root-finding taylorinteg has no dense output option.

Many thanks for this change!

PerezHz:
> Currently, propagate_root cannot undergo the same change since root-finding taylorinteg has no dense output option.

Then maybe I could update https://github.com/PerezHz/TaylorIntegration.jl/pull/168 to include the missing feature? What do you guys think?

PerezHz:
Heads up: I pushed a couple of commits updating the artifacts

PerezHz:
It seems the MPC website is not fond of receiving requests from `Downloads.curl`... I'll try to take a look into this

PerezHz:
Just retriggered CI... good news is, when the download issues don't get in the way, tests actually pass!

lbenet:
> Then maybe I could update https://github.com/PerezHz/TaylorIntegration.jl/pull/168 to include the missing feature? What do you guys think?

While I agree to update https://github.com/PerezHz/TaylorIntegration.jl/pull/168, what do you have in mind? More concretely, the question is: which time shall we associate with the (dense output) polynomials, the crossing time (actual Poincare surface crossing, or root of the function `g`), or the previous time, which includes the crossing as well? 

Note that when doing jet transport, the time of crossing is actually a polynomial in the jet transport coordinates.

In any case, if you have a concrete idea, go ahead a push a commit...

PerezHz:
Sorry, I pushed two commits which I meant to push to another branch; it's now fixed... 😓

PerezHz:
> While I agree to update https://github.com/PerezHz/TaylorIntegration.jl/pull/168, what do you have in mind?

Since the root-finding times are well-defined without the dense output, for the root-finding method with dense output, my first idea for root-finding with dense ouput would be to follow as closely as possible the root-finding method of `taylorinteg` without dense output. But I'm not sure if that's consistent with `propagate_root` here...

PerezHz:
I ran the tests from this branch of NEOs.jl together with the PerezHz/TaylorIntegration.jl#168 branch, and I got some errors in PE due to some assertions in interpolation. Those can be fixed quick, but maybe it makes more sense to actually merge the PR#168 in TI essentially as it is now, and then patch things if/when necessary. This way we can track compatibility, etc., with CI. Regarding the `taylorinteg`  method with root-finding and dense output, I can take care of that in a separate PR in TI.

lbenet:
I agree; go ahead and merge https://github.com/PerezHz/TaylorIntegration.jl/pull/168, and either we skip that test in Julia nightly, of somehow fix it.

PerezHz:
I think the only thing pending here is to add the root-finding with dense ouput ` taylorinteg` method instead of calling `neosinteg` in ` propagate_root` ; I'm on it and will hopefully have some time later today to push a new PR in TaylorIntegration

PerezHz:
Update: [PerezHz/TaylorIntegration#170](https://github.com/PerezHz/TaylorIntegration.jl/pull/170) has been merged and TI v0.14.1 has been released; I updated this PR with the corresponding changes, the most important one being `neosinteg` is no more... Time for Saturn to eat his children I guess 😅

PerezHz:
This is now ready for review @LuEdRaMo @lbenet

lbenet (commented):
Thanks @PerezHz for this PR. In general things LGTM.

While we have green lights, I wonder if you have checked some results (e.g., Apophis) using this PR in comparison to the old ones? I'm in favor of merging this, but prefer to do it once we have an independent check.

PerezHz:
> I wonder if you have checked some results (e.g., Apophis) using this PR in comparison to the old ones? I'm in favor of merging this, but prefer to do it once we have an independent check.

I agree that we should have an independent check; I'll run this PR on Apophis and will report back!

PerezHz:
Update: I removed now-unused the non-threaded asteroid dynamical model, which was really useful back in the day when to debug the multi-threaded version. Another change I did was to transition from `SatelliteToolbox` to `SatelliteToolboxTransformations`, a lightweight dependency which has the whole functionality we actually need for the ECEF <-> ECI transformations. I can report that offline tests show that this PR together with the latest releases of PE and TI can reproduce the results from our paper, including the Yarkovsky transverse parameter A2 together with the associated uncertainties and post-fit statistics (normalized RMS, chi-squared, etc.). There are some minor updates pending related to some scripts and notebooks, but all in all I think we can go ahead and merge this PR, since the update to PE v0.7 is complete at last!

PR #29, opened 2023-07-11
Update Apophis script and some cleanup

### LINKED ISSUE #20: Recurrent failure of observations tests due to timeout error in Windows
Not terribly urgent, but fixing this issue would allow us to avoid re-run tests which would otherwise pass

PerezHz:
This PR updates the `pha/apophis.jl` script, and adds a `Project.toml` in the same `pha` folder to execute the script independently of the NEOs.jl environment. I've also taken the opportunity to do cleanup of some legacy code, and tried to fix #20 by going back to downloading MPC catalogue codes via HTTP.jl (similar to how it was originally done). Yet, there are still some issues related to windows testing of NEOs.jl via GHA CI, for which I don't have a quick fix right now.


PerezHz:
Tests are failing in nightly due to some issues with StaticArrays, but I suggest to wait a bit and see before taking action on this. Tests are also failing on windows, but that seems to be CI-related, since offline tests pass without any issue, as far as we've been able to test on different windows machines. Therefore, since we have passing tests on ubuntu and macos, I will go ahead and merge.

PR #31, opened 2023-07-17
Initial Orbit Determination

LuEdRaMo:
This PR introduces a new function `gaussinitcond` which transforms a vector of optical observations into a vector of initial conditions via `gauss_method`.


PerezHz (commented):
Many thanks for this contribution @LuEdRaMo! This is an awesome contribution to the preliminary OD capabilities in NEOs.jl! Not sure if this is ready for review, but left a couple of comments around; hope this helps!

ext/DataFramesExt.jl
```diff
@@ -30,4 +36,158 @@ end
...
+- `maxsteps::Int`: maximum number of steps for propagation.
+- `varorder::Int`: order of jet transport perturbation. 
+- `order::Int`: order of Taylor polynomials w.r.t. time.
+- `abstol::T`: absolute tolerance.
+- `parse_eqs::Bool`: whether to use the taylorized method of `RNp1BP_pN_A_J23E_J2S_eph_threads!` or not. 
+
+!!! warning
+    This function will set the (global) `TaylorSeries` variables to `δα₁ δα₂ δα₃ δδ₁ δδ₂ δδ₃`. 
+"""
+function gaussinitcond(radec::Vector{RadecMPC{T}}; Δ::DatePeriod = Day(1), niter::Int = 5, maxsteps::Int = 100, varorder::Int = 5,
```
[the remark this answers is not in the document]
LuEdRaMo: `gaussinitcond` depends on `reduce_nights` and the latter depends on functions that only exist within the `DataFrames` extension, so I'm not sure if `gaussinitcond` would work on `src/`.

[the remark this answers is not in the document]
PerezHz: Ok, that's a good point, but then this functionality is only available when `DataFrames` is loaded, right? If the `DataFrames` dependency turns out to be a not-so-weak dep, then maybe it makes sense to add it as a full dep?

src/orbit_determination/gauss_method.jl
```diff
@@ -310,4 +310,83 @@ function gauss_method(observatories::Vector{ObservatoryMPC{T}}, dates::Vector{Da
...
-end
\ No newline at end of file
+end
+
+@doc raw"""
+    vandermonde(x::AbstractVector{T}, order::Int) where {T <: Real}
+
+Return the Vandermonde matrix of order `order` for points `x`. 
+"""
+function vandermonde(x::AbstractVector{T}, order::Int) where {T <: Real}
```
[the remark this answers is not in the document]
LuEdRaMo: Actually, I've been doing some tests and it turns out that polynomial interpolation is not very stable as the number of points increases. I think that in the next commit I will switch to the `Interpolations` package.

PerezHz:
@LuEdRaMo many thanks for all the effort here! I'm seeing lots of files with only white-space changes, can you please revert those?

LuEdRaMo:
> @LuEdRaMo many thanks for all the effort here! I'm seeing lots of files with only white-space changes, can you please revert those?

Hi @PerezHz, those changes must have happened either when I changed laptop or when I rebased to main. Anyway, those changes are gone.

PerezHz:
> Hi @PerezHz, those changes must have happened either when I changed laptop or when I rebased to main. Anyway, those changes are gone.

Thanks a lot! This is just to help keep the file history a bit cleaner. One more thing: there's a `pha/.DS_Store` binary file added as part of this PR, is this actually needed?

LuEdRaMo:
> Thanks a lot! This is just to help keep the file history a bit cleaner. One more thing: there's a `pha/.DS_Store` binary file added as part of this PR, is this actually needed?

It is a macOS internal file that slipped through, I've deleted it.

src/propagation/parameters.jl
```diff
@@ -0,0 +1,96 @@
+struct Parameters{T <: AbstractFloat}
```
PerezHz, line 1: I feel we can make the name `Parameters` a bit more descriptive, maybe `NEOParameters`, `NEOParameters`?

src/propagation/asteroid_dynamical_models.jl
```diff
@@ -1,1176 +1,1176 @@
...
+    ζz2 = Array{S}(undef, N)
+
+    # Full extended-body accelerations
+    accX = zero_q_1
+    accY = zero_q_1
+    accZ = zero_q_1
+
+    # Rotations to and from Earth, Sun and Moon pole-oriented frames
+    local M_ = Array{S}(undef, 3, 3, N)
+    local M_ .= zero_q_1
```
PerezHz, line 309: great catch!

src/postprocessing/least_squares.jl
```diff
@@ -1,437 +1,681 @@
...
-    end
-    # Normal matrix
-    C = TaylorSeries.hessian(Q, x_new)/(2/nobs) # C = d2Q/(2/m)
-    # Covariance matrix
-    Γ = inv(C)
-
-    return x_new, Γ, C
-end
+@doc raw"""
+    OrbitFit{T <: Real}
```
PerezHz, line 2: If `OrbitFit` represents a least squares fit, then I suggest to rename it to `LeastSquaresFit`. Also, it can be argued that these sub-routines belong conceptually more to orbit determination, so I also suggest to move this file to the `orbit_determination` folder.

test/orbit_determination.jl
```diff
@@ -0,0 +1,143 @@
...
+        @test iszero(count(outlier.(sol.res)))
+        # Least squares fit
+        @test sol.fit.success
+        @test all( sigmas(sol) .< 1e-4 )
+        @test nrms(sol) < 0.38
+        # Scalig factors
+        @test all(sol.scalings .== 1e-6)
+    end
+
+    @testset "Too Short Arc" begin
```
PerezHz, line 51: Thanks for adding these tests! Have you tried these subroutines with objects such as 2008TC3 and 2014AA? I think those would be ideal test cases...

src/observations/topocentric.jl
```diff
@@ -1,151 +1,318 @@
...
+    sunriseset(date::DateTime, observatory::ObservatoryMPC{T}) where {T <: AbstractFloat}
+    sunriseset(radec::RadecMPC{T}) where {T <: AbstractFloat}
+
+Return `DateTime` of sunrise and sunset at a particular date and location.
+
+!!! reference
+    See "General Solar Position Calculations" by NOAA at
+    https://gml.noaa.gov/grad/solcalc/solareqns.PDF.
+"""
+function sunriseset(date::DateTime, observatory::ObservatoryMPC{T}) where {T <: AbstractFloat}
```
PerezHz, line 94: We have the observer's position in ECEF frame as given from the MPC data, and we also know the Sun's ephemeris which we can transform to ECEF frame. Then, it's straight-forward to compute the elevation angle as a function of time, and from there it's straight-forward to compute the sunrise and sunset via root-finding, including possibly the offset due to the solar disk. I think we can get a more accurate computation from there?

src/propagation/propagation.jl
```diff
@@ -44,6 +42,26 @@ function scaled_variables(names::String = "δx", c::Vector{T} = fill(1e-6, 6); o
...
+        c = sol.t[end] < t
+    elseif issorted(sol.t, rev = true)
+        c = sol.t[end] > t
+    else
+        c = true
+    end
+
+    return iszero(sol) || any(norm.(sol.x, Inf) .> 10) || c 
+end
+
```
PerezHz, line 64: Since this is a predicate, I would suggest to rename it to something such as `issuccessfulprop` (and changing the logic appropriately)

test/orbit_determination.jl
LuEdRaMo: Thank you for the suggestion, both are indeed very interesting asteroids. Currently,  2014 AA converges to an orbit without a problem via `orbitdetermination`. On the other hand, the routine is not well suited for an object like 2008 TC3 that has a lot of observations covering a relatively short time. In future PRs I will address performance and converge issues like this.

src/observations/topocentric.jl
LuEdRaMo: That's a great suggestion and most likely would improve the accuracy of sunrise(set) times. However, this function only purpose is to group observations by night and that division is usually very neat, so we don't really need a lot of precision in that calculation.  I am in favor of leaving this as it is unless we find a specific asteroid where such a precision makes a difference in the way we group observations.

src/postprocessing/least_squares.jl
LuEdRaMo: I agree that `LeastSquaresFit` is a better name, so I've changed all `OrbitFit` to `LeastSquaresFit` and moved the files accordingly.

src/propagation/parameters.jl
LuEdRaMo: Since the struct contains parameters used for many routines apart from propagation, I've decide to change all `Parameters` to `NEOParameters`.

test/orbit_determination.jl
PerezHz: > Currently, 2014 AA converges to an orbit without a problem via orbitdetermination. 

Nice, then I suggest to add that 2014 AA to the tests!

> On the other hand, the routine is very slow for an object like 2008 TC3 that has > 800 observations spanning < 1 day

What happens if you try doing the same but using a tracklet with only the first four observations from Mt. Lemmon in the MPC astrometry file for 2008 TC3?

src/observations/topocentric.jl
PerezHz: I'm wondering what happens if there's some observations near the local sunrise/sunset, but miss the grouping by, say, one minute, the it gets grouped in the next (or previous) day? Anyway, since this more an accuracy enhancement for edge cases rather than an fix, then we can probably go ahead with the method you propose, and keeping in mind an improvement if it's needed after merging this PR.

src/observations/observation_night.jl
```diff
@@ -0,0 +1,182 @@
...
+- `date::DateTime`: mean date of observation.
+- `α::T`: mean right ascension.
+- `δ::T`: mean declination.
+- `v_α::T`: mean right ascension velocity.
+- `v_δ::T`: mean declination velocity.
+- `mag::T`: mean apparent magnitude.
+- `nobs::Int`: number of observations.
+- `idxs::Vector{Int}`: indices of original `radec` that entered the night.
+"""
+@auto_hash_equals struct ObservationNight{T <: AbstractFloat}
```
PerezHz, line 20: After giving it some thought, I think what this `struct` represents, matches very well with what the notion of "tracklet" means in NEO literature (i.e., the group of observations for a given object, during a single night, from a single observatory), so I suggest to rename this `struct` to `Tracklet`.

src/propagation/propagation.jl
LuEdRaMo: I've changed all `unsuccessful_propagation` to `issuccessfulprop`, as well as the corresponding logic.

src/observations/observation_night.jl
PerezHz: Another possibility is to follow Milani's `VeryShortArc` terminology, although my impression is that `Tracklet` is more common these days, so I'd be more inclined to the latter...

LuEdRaMo: I agree that both names are better than `ObservationNight`; since `Tracklet` is shorter, I've changed all corresponding names to `Tracklet`.

src/observations/topocentric.jl
LuEdRaMo: I've been testing the OD routines with a 5,000 NEOs sample. I will check if such scenario occurs in the sample. If that is the case, we can change to a more accurate method in a future PR.

test/orbit_determination.jl
LuEdRaMo: > Nice, then I suggest to add that 2014 AA to the tests!

2014 AA is now part of the tests.

> What happens if you try doing the same but using a tracklet with only the first four observations from Mt. Lemmon in the MPC astrometry file for 2008 TC3?

It is posible to get an orbit using `orbitdetermination` with the 15 observations before 6/Oct/08 10:00 UTC and then add the remaining observations via `propres` + `tryls`. I've also added this handcrafted OD solution to the tests.

PerezHz: Awesome, thanks for adding these tests!

src/orbit_determination/tooshortarc.jl
```diff
@@ -0,0 +1,549 @@
...
+    v_ρ_domain = [v_ρ_min, v_ρ_max]
+    # Range rate symmetry level
+    v_ρ_mid = range_rate(coeffs, ρ_max)[1]
+    # Boundary points
+    Fs = Matrix{Float64}(undef, 3, 2)
+    Fs[1, :] .= [ρ_min, v_ρ_min]
+    Fs[2, :] .= [ρ_min, v_ρ_max]
+    Fs[3, :] .= [ρ_max, v_ρ_mid]
+    
+    return AdmissibleRegion{Float64}(t_datetime, α, δ, v_α, v_δ, ρ, ρ_α, ρ_δ, q,
```
PerezHz, line 91: Since the method has the type parameter `T<:AbstractFloat` I suggest to also use it for the returned value
```suggestion
    return AdmissibleRegion{T}(t_datetime, α, δ, v_α, v_δ, ρ, ρ_α, ρ_δ, q,
```

src/observations/tracklet.jl
```diff
@@ -0,0 +1,182 @@
...
+
+Return the coefficients of the derivative of a polynomial with coefficients `x`.
+"""
+function diffcoeffs(x::Vector{T}) where {T <: AbstractFloat}
+    y = Vector{T}(undef, length(x)-1)
+    for i in eachindex(y)
+        y[i] = i * x[i+1]
+    end
+    return y
+end
```
PerezHz, line 99: Have you considered using Polynomials.jl for this functionality? If Polynomials.jl has similar functionalities (fitting a polynomial model to an univariate time series) it's probably worth it giving it a try

test/orbit_determination.jl
```diff
@@ -0,0 +1,267 @@
...
+        # Optical astrometry file
+        filename = joinpath("data", "2008_EK68.txt")
+        # Download observations
+        get_radec_mpc("designation" => "2008 EK68", filename)
+        # Load observations 
+        radec = read_radec_mpc(filename)
+        # Delete astrometry file
+        rm(filename)
```
PerezHz, line 65: In some of the tests there seems to be a pattern of opening a file, constructing a `Vector{RadecMPC{T}}` and then deleting the file. I suggest to add a function like this, which can reduce this pattern to a single line
```julia
function fetch_radec_mpc(s::String)
    f = tempname()
    get_radec_mpc("designation" => "2008 EK68", f)
    radec = read_radec_mpc(filename)
    rm(f)
    return radec
end
```

PerezHz, line 65: EDIT: (deleted my previous suggestion, since it had a typo 😅)

In some of the tests there seems to be a pattern of opening a file, constructing a Vector{RadecMPC{T}} and then deleting the file. I suggest to add a function like this, which can reduce this pattern to a single line

```julia
function fetch_radec_mpc(s::String)
    f = tempname()
    get_radec_mpc("designation" => s, f)
    radec = read_radec_mpc(filename)
    rm(f)
    return radec
end
```

src/orbit_determination/tooshortarc.jl
LuEdRaMo: Done.

test/orbit_determination.jl
LuEdRaMo: Thanks for the suggestion, I've added `fetch_radec_mpc` to `src/observations/radec_mpc.jl` and modified the tests accordingly.

src/observations/tracklet.jl
LuEdRaMo: I think it's better to have `LsqFit.jl` as a dependency, in case we ever need to fit a function other than a polynomial. For instance, a future implementation of ADS (see https://github.com/PerezHz/TaylorIntegration.jl/pull/178) might require fitting exponentials.

src/NEOs.jl
```diff
@@ -1,75 +1,84 @@
...
-
+module NEOs
+
+# __precompile__(false)
+
+import Base: show, string, hash, isequal, ==, isless, convert, zero, iszero, isnan, in
+import Tables: istable, rowaccess, rows, schema, Schema
+import SatelliteToolboxTransformations: sv_ecef_to_eci, sv_ecef_to_ecef
+import JLD2: writeas
+import PlanetaryEphemeris as PE
```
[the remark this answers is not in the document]
LuEdRaMo: Great catch! I've changed the `Project.toml` julia entry to 1.9.

src/observations/topocentric.jl
```diff
@@ -1,151 +1,337 @@
...
+- `start::DateTime`.
+- `stop::DateTime`.
+- `utc::Int`: hours from UTC.
+"""
+@auto_hash_equals struct TimeOfDay
+    light::Symbol
+    start::DateTime
+    stop::DateTime
+    utc::Int
+    function TimeOfDay(date::DateTime, observatory::ObservatoryMPC{T}) where {T <: AbstractFloat}
```
lbenet, line 19: I'm checking (slowly) the code, but I wonder about the need of this function (from the docstrings). My point is the following: If we know the observatory (and assuming an optical observation), then (1) the observatory is Earth-based, in which case the observation did take place at "local night"; (2) the observation took place from the space, in which case we don't care about the local night or day time.
What am I missing?

LuEdRaMo: You're right, in all space-based observations we do not care about the day/night, and all Earth-based observations occur at "local night". However, we must compare observations from different stations and times. To do so, we must know the UTC time of "dusk" and "dawn" at each night.

src/propagation/parameters.jl
```diff
@@ -0,0 +1,113 @@
...
+                            resol, bias_matrix, eph_su, eph_ea, niter, max_triplets,
+                            varorder, gaussQmax, H_max, maxiter, tsaQmax, max_per)
+end
+
+function NEOParameters(params::NEOParameters{T}; kwargs...) where {T <: AbstractFloat}
+    fields = fieldnames(NEOParameters{T})
+    vals = Vector{Any}(undef, length(fields))
+    for i in eachindex(vals)
+        if fields[i] in keys(kwargs)
+            vals[i] = kwargs[i]
```
PerezHz, line 101: If I do
```julia
using NEOs
params = NEOParameters()
params = NEOParameters(params, parse_eqs=false)
```
then I get an error; one way to solve this is as follows:

```suggestion
            vals[i] = kwargs[fields[i]]
```

LuEdRaMo: This was fixed by https://github.com/PerezHz/NEOs.jl/pull/31/commits/e8bf9bb596236269685ab95f01e486a8ccf69003.

PerezHz:
Just pushed a fix for some tests after updating to PE v0.8.1, but feel free to update them if you see fit 👍

PerezHz:
(I actually meant to push it to another branch, let me know if you'd rather to grab it from a PR)

LuEdRaMo:
> (I actually meant to push it to another branch, let me know if you'd rather to grab it from a PR)

Feel free to push on this branch.

src/orbit_determination/gauss_method.jl
```diff
@@ -114,125 +118,142 @@ end
...
+    solve_lagrange(a::TaylorN{T}, b::TaylorN{T}, c::TaylorN{T}; niter::Int = 5) where {T <: Real}
 
 Solve Lagrange polynomial.
 
 See also [`lagrange`](@ref).
 """
 function solve_lagrange(a::T, b::T, c::T; niter::Int = 5) where {T <: Real}
-    sol = roots(x -> lagrange(x, a, b, c), Interval(0.00465047, 40))
-    return mid.(interval.(sol)), getfield.(sol, :status)
+    return find_zeros(x -> lagrange(x, a, b, c), 0.00465047, 40.0)
```
PerezHz, line 153: I suggest to add the boundaries to the search intervals as keyword arguments to `solve_lagrange` instead of having them set manually in the function body

src/orbit_determination/gauss_method.jl
```diff
@@ -301,13 +324,225 @@ function gauss_method(observatories::Vector{ObservatoryMPC{T}}, dates::Vector{Da
...
+    sort!(triplets, by = x -> gauss_norm(dates[x]))
+
+    n = min(length(triplets), max_triplets)
+
+    return triplets[1:n]
+end
+
+function gauss_triplets(dates::Vector{DateTime}, max_triplets::Int = 10, max_iter::Int = 100)
+    Δ_min = Hour(20)
+    Δ_max = Day(7)
```
PerezHz, line 398: I'd suggest to keep `Δ_min` and `Δ_max` as keyword arguments in `gauss_triplets`, with default values as they are now

src/orbit_determination/tooshortarc.jl
```diff
@@ -0,0 +1,718 @@
...
+    scalings[1:4] .= LOS ./ 1e6
+    if scale == :linear
+        scalings[5] = (A.ρ_domain[2] - A.ρ_domain[1]) / 1_000
+    elseif scale == :log
+        x = log10(ρ)
+        scalings[5] = (log10(A.ρ_domain[2]) - log10(A.ρ_domain[1])) / 1_000
+    end
+    scalings[6] = (A.v_ρ_domain[2] - A.v_ρ_domain[1]) / 1_000
+    # Jet transport variables
+    dq = scaled_variables("dx", scalings, order = 2)
```
PerezHz, line 431: I'd suggest to keep `order` as a keyword argument to `adam`

src/orbit_determination/tooshortarc.jl
```diff
@@ -0,0 +1,718 @@
...
+               i::Int, ρ::T, v_ρ::T, params::NEOParameters{T}; maxiter::Int = 5, dynamics::D=newtonian!) where {T <: AbstractFloat, D}
+    # Initial time of integration [julian days]
+    # (corrected for light-time)
+    jd0 = datetime2julian(A.date) - ρ / c_au_per_day
+    # Barycentric initial conditions
+    q0 = topo2bary(A, ρ, v_ρ)
+    # Scaling factors
+    scalings = abs.(q0) ./ 10^5
+    # Jet transport variables
+    dq = scaled_variables("dx", scalings, order = 6)
```
PerezHz, line 555: Similar to previous comments, I'd suggest to add `order` as a keyword argument to `tsals`, maybe with a different name (`varorder`?) and set `order=varorder` here

src/observations/topocentric.jl
PerezHz: Probably worth to add something to the docstrings about this point, something siimilar to what you said earlier: this is used to decide whether a number of observations correspond to the same observation batch, taken at the same time during the same night (or something similar to this).

src/orbit_determination/gauss_method.jl
LuEdRaMo: Solved by https://github.com/PerezHz/NEOs.jl/pull/31/commits/b67b85ec7f3e4b5d0bf443cce1388194d6fbf464.

src/orbit_determination/gauss_method.jl
LuEdRaMo: Solved by https://github.com/PerezHz/NEOs.jl/pull/31/commits/b67b85ec7f3e4b5d0bf443cce1388194d6fbf464.

src/orbit_determination/tooshortarc.jl
LuEdRaMo: Solved by https://github.com/PerezHz/NEOs.jl/pull/31/commits/b67b85ec7f3e4b5d0bf443cce1388194d6fbf464.

src/orbit_determination/tooshortarc.jl
LuEdRaMo: Solved by https://github.com/PerezHz/NEOs.jl/pull/31/commits/b67b85ec7f3e4b5d0bf443cce1388194d6fbf464.

LuEdRaMo:
@PerezHz, @lbenet, I've pushed the changes we discussed in our last meeting. Let me know if you have more comments...

src/propagation/parameters.jl
```diff
@@ -0,0 +1,136 @@
...
+    for i in eachindex(vals)
+        if fields[i] in keys(kwargs)
+            vals[i] = kwargs[fields[i]]
+        else
+            vals[i] = getfield(params, i)
+        end
+    end
+
+    return NEOParameters{T}(vals...)
+end
```
PerezHz, line 131: (Parameters.jl)[https://github.com/mauro3/Parameters.jl] and `Base.@kwdef` implement similar functionality, have you considered using either of those to implement that funcionaility here?

PerezHz:
@LuEdRaMo thank you for adding the suggestions we discussed offline; I felt a couple of comments but I aside from those I think this PR is pretty much ready to be merged; any further improvements we can always address in another PR. This has been really an awesome contribution, truly many thanks!

test/orbit_determination.jl
```diff
@@ -0,0 +1,301 @@
...
+        @test length(sol1.res) == 30
+        @test iszero(count(outlier.(sol1.res)))
+        # Least squares fit
+        @test sol1.fit.success
+        @test all( sigmas(sol1) .< 2e-6 )
+        @test nrms(sol1) < 0.37
+        # Scalig factors
+        @test all(sol1.scalings .< 1e-6)
+        # Compatibility with JPL
+        @test all(abs.(sol1() - JPL) ./ sigmas(sol1) .< 1.6)
```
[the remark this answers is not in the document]
LuEdRaMo: JPL gives no information about which observations they discard as outliers or which weights they use; therefore, a strict comparison with JPL is not possible. Furthermore, the first orbit is well constrained because it is computed over the best quality astrometry, the observations added for the second orbit do not span a significant amount of extra time and may only introduce noise and/or outliers. Yet, the comparison stays within a 3 sigma threshold wrt JPL. This particular NEO requires a detailed filtering of the observations.

src/propagation/parameters.jl
LuEdRaMo: I implemented this manually to have more flexibility with `NEOParameters`, specially wrt the debiasing tables. The user specifies the table as a string (e.g. `debias_table = "2018"`) but the struct saves the corresponding variables (a string, a `Healpix` varible and a matrix), and I do not know if that is possible with the options you mention.

test/orbit_determination.jl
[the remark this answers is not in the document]
PerezHz: Okay, if comparison wrt a JPL solution is not straightforward, then probably it makes sense to test for improvement of the post-fit statistics for each solution, e.g.:
```julia
@test nrms(sol1) < nrms(sol)
@test all.(sigmas(sol1) .< sigmas(sol))
```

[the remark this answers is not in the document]
LuEdRaMo: The NRMS increases from 0.29 to 0.37 from `sol` to `sol1` but `sigmas(sol1)` are smaller than `sigmas(sol)` by a factor of 4. I've added the test
```julia
@test all(sigmas(sol1) .< sigmas(sol))
```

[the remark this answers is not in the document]
PerezHz: Alright, then I'd only suggest now to add the NRMS test commented for now and leave as a TODO to try to understand a bit better the differences wrt to the JPL solution.

test/orbit_determination.jl
```diff
@@ -0,0 +1,303 @@
...
+        # 2008 TC3 entered the Earth's atmosphere around October 7, 2008, 02:46 UTC
+
+        # Fetch optical astrometry
+        radec = fetch_radec_mpc("designation" => "2008 TC3")
+
+        # Parameters
+        params = NEOParameters(coeffstol = Inf, bwdoffset = 0.007, fwdoffset = 0.007)
+
+        # Observations with <1" weight
+        idxs = findall(x -> x < 1, w8sveres17.(radec))
```
PerezHz, line 232:
```suggestion
        idxs = findall(x -> w8sveres17(x) < 1, radec)
```
(avoids creating an intermediate array)

test/orbit_determination.jl
```diff
@@ -0,0 +1,303 @@
...
+        @test all( norm.(sol.fwd.x, Inf) .< 1e4 )
+        @test isempty(sol.t_fwd) && isempty(sol.x_fwd) && isempty(sol.g_fwd)
+        # Vector of residuals
+        @test length(sol.res) == 20
+        @test iszero(count(outlier.(sol.res)))
+        # Least squares fit
+        @test sol.fit.success
+        @test all( sigmas(sol) .< 2e-5 )
+        @test nrms(sol) < 0.30
+        # Scalig factors
```
PerezHz, line 263:
```suggestion
        # Scaling factors
```
(there's a missing n in `Scaling`)

test/orbit_determination.jl
LuEdRaMo: Done.

test/orbit_determination.jl
LuEdRaMo: Thank you for noting that. I've corrected all `Scaling` misspellings in that file.

PerezHz:
Just one minor detail: since the DataFrames extension is not a weakdep anymore, I'd suggest to rename `test/extensions.jl` to e.g. `test/dataframe.jl`.

LuEdRaMo:
> Just one minor detail: since the DataFrames extension is not a weakdep anymore, I'd suggest to rename `test/extensions.jl` to e.g. `test/dataframe.jl`.

Done.

lbenet:
Excellent! Thanks a lot @LuEdRaMo

PR #53, opened 2024-02-06
Add Newtonian model as default dynamics for preliminary orbit determination

PerezHz:
@LuEdRaMo this can help performance of #31 when not so much fidelity is needed in the dynamics.


LuEdRaMo:
Thanks a lot @PerezHz! This is a massive gain in performance while keeping the OD results consistent. I'll merge and keep working on this in https://github.com/PerezHz/NEOs.jl/pull/31.

Issue #30, opened 2023-07-11
Avoid type piracy of `SatelliteToolboxTransformations.jl` overloads


PerezHz:
Currently, here:
https://github.com/PerezHz/NEOs.jl/blob/7e0ceec3e2eeea03679d9b8140432baeeafd09a0/src/observations/topocentric.jl#L42
and here:
https://github.com/PerezHz/NEOs.jl/blob/7e0ceec3e2eeea03679d9b8140432baeeafd09a0/src/observations/topocentric.jl#L59
we are overloading methods defined in `SatelliteToolboxTransformations.jl` to be able to differentiate through those functions using `TaylorSeries` `Taylor1` and `TaylorN`; which is type piracy. It's not terrible, but I think we should try to avoid it. AFAIU, this is due to products of `StaticArray.SMatrix` with vectors of `<:AbstractSeries` sometimes giving `Vector{Any}` as a result, which is why the results of such products are type annotated in the overloaded methods here. Probably this issue belongs in `TaylorSeries`, but thought of opening this issue here to discuss and see how we might solve this.

lbenet:
A way to overcome this (weak) type piracy would be to pass a `TaylorInterpolant`... Does this make sense?

Incidentally, have you looked to `Aqua.jl`? It may be worth...

PerezHz:
Solved by JuliaDiff/TaylorSeries.jl#353

PR #62, opened 2024-03-26
Test with Aqua.jl

lbenet:


lbenet:
This PR aims to check with Aqua.jl, specially for ambiguities, and issues with type piracy, which may affect performance

lbenet:
Aqua reports some [ambiguities](https://github.com/PerezHz/NEOs.jl/actions/runs/8443150402/job/23126143664?pr=62#step:6:287), and possible [test piracies](https://github.com/PerezHz/NEOs.jl/actions/runs/8443012596/job/23125811505?pr=62#step:6:311) (only in Julia v1.10). Note that I have commented several tests to see the output from Aqua (which strangely differs from what I get locally).

PerezHz:
Many thanks for this PR @lbenet! Indeed, there are some type piracies related to some overloads of SatelliteToolboxTransformations methods (reported in #30), then for the time being I think it's okay to mark those tests as broken as you've already done.

PerezHz:
Regarding the method ambiguities I think those come from Interpolations.jl, it's probably worth opening an issue there?

lbenet:
Rebased to current master...

Project.toml
```diff
@@ -56,4 +61,5 @@ StatsBase = "0.33, 0.34"
 Tables = "1.10"
 TaylorIntegration = "0.15"
 TaylorSeries = "0.17"
+Test = "1"
```
PerezHz, line 64: `Test` is no longer included in NEOs deps, so I think it's safe to remove this compat entry
```suggestion
```

Project.toml
```diff
@@ -37,16 +37,21 @@ Artifacts = "1"
 AutoHashEquals = "0.2"
 Clustering = "0.15"
 DataFrames = "1.5"
+Dates = "1"
 DelimitedFiles = "1"
+Distributed = "1"
```
PerezHz, line 42:
```suggestion
```

PerezHz:
(I went ahead with a couple of commits to remove Test and Distributed compat entries)

PerezHz:
Also, just opened https://github.com/JuliaSpace/SatelliteToolboxTransformations.jl/issues/6 which I hope can shed some light into how to avoid overloading `ecef_to_geocentric` for `Vector{<:AbstractSeries}`

PerezHz:
Pushed a few commits which, together with JuliaDiff/TaylorSeries.jl#353, solve method ambiguities and a possible type piracy

PerezHz:
JuliaDiff/TaylorSeries.jl#354 solves another type piracy, but that's not strictly necessary as things are now, since piracy tests are marked as broken (we can either solve the piracies or leave those tests marked as broken for the time being).

src/observations/process_radec.jl
```diff
@@ -44,7 +44,7 @@ function evaluate(res::AbstractVector{OpticalResidual{T, TaylorN{T}}}, x::Vector
     end
     return res_new
 end
-(res::AbstractVector{OpticalResidual{T, TaylorN{T}}})(x::Vector{T}) where {T <: Real} = evaluate(res, x)
+(res::Vector{OpticalResidual{T, TaylorN{T}}})(x::Vector{T}) where {T <: Real} = evaluate(res, x)
```
lbenet, line 47: Naive question: what is the motivation for this change? (I think `AbstractVector` would allow applying this function to more types, e.g. `SubArray{T,1}` or `SVector`).

lbenet: This may perhaps not be relevant, but I thought it was easier to ask...

PerezHz: This helps avoid an ambiguity with callable methods from Interpolations.jl which also dispatch on `<:AbstractArray{T,N}`; I agree it's perhaps a bit restrictive, and a more generic solution could be to define a `OpticalResidualVector` struct and define the callable methods over that struct. I would suggest to go with this approach for now, and revisit this if needed.

PerezHz:
Pushed a couple of commits reflecting the current state of things after merging of JuliaDiff/TaylorSeries.jl#354

PerezHz:
There are still some type-piracies due to some TaylorInterpolant JLD2 serialization dispatches; since TaylorInterpolant lives in PlanetaryEphemeris that type-piracy can be solved by moving over those dispatches over to PlanetaryEphemeris (working on it)

src/observations/process_radec.jl
PerezHz: Yet another possibility is to avoid using `Vector{T}` for the calling argument type (again, maybe wrapping it somehow in another struct)

PerezHz:
Solved the remaining type-piracies by moving over the corresponding methods to PlanetaryEphemeris. Are there any other thing you think we should add here @lbenet?

lbenet:
Thanks a lot @PerezHz for solving all the remaining issues related to this PR. I am in favor of merging it as it is.

PR #65, opened 2024-04-10
`RecipesBase` extension

LuEdRaMo:
This PR introduces a `RecipesBase` extension with custom recipes to plot some of `NEOs`' native objects. So far, there are only two recipes: `AbstractVector{OpticalResidual}` and `AdmissibleRegion`, although suggestions are welcomed @lbenet @PerezHz .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NEOs"
 uuid = "b41c07a2-2abb-11e9-070a-c3c1b239e7df"
 authors = ["Jorge A. Pérez Hernández", "Luis Benet", "Luis Eduardo Ramírez Montoya"]
-version = "0.8.3"
+version = "0.8.4"
 
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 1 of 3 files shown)


PerezHz:
Hey @LuEdRaMo I think this is a really valuable addition to the repo! Besides the recipes you added, adding a simple orbit viewer and a visualization of the uncertainty region projection onto a close approach B-plane could be really useful.

LuEdRaMo:
> Hey @LuEdRaMo I think this is a really valuable addition to the repo! Besides the recipes you added, adding a simple orbit viewer and a visualization of the projection of the uncertainty region onto the B-plane during a close approach can be really useful.

I added a simple orbit viewer as you suggested. However, to have a recipe for the uncertainty region in the B-plane we would need a custom NEOs struct and I think that should be the subject of another PR.

PerezHz:
Thanks for adding the orbit viewer! And sure, we can discuss adding a NEO struct in a future PR. Can you add a simple example of how to use the extension for orbit and optical residuals plotting? Maybe add a small example in a .jl file in a new `examples` folder for a given NEO.

LuEdRaMo:
> Thanks for adding the orbit viewer! And sure, we can discuss adding a NEO struct in a future PR. Can you add a simple example of how to use the extension for orbit and optical residuals plotting? Maybe add a small example in a .jl file in a new `examples` folder for a given NEO.

I've added a file `examples/recipes.jl` illustrating the three recipes we have so far with asteroid 2023 DW.

LuEdRaMo:
If you have no further suggestions @PerezHz @lbenet, I think this is ready to merge. If we need more recipes, we can add them in future PRs.

lbenet (approved):
Thanks @LuEdRaMo for this addition, including the example. LGTM. So, go ahead, merge and tag the new version

PR #66, opened 2024-04-18
Curvature computations according to Milani et al. (2008)

LuEdRaMo:
The curvature of an arc is fundamental to asses the reliability of an orbit. Therefore, this PR introduces the function `curvature` to compute the geodesic curvature and along track acceleration of a vector of optical observations, as well as its covariance matrix, following [Milani et al. (2008)](https://doi.org/10.1016/j.icarus.2007.11.033).


LuEdRaMo:
This PR is ready. Since tests pass I'll proceed to merge.

PR #67, opened 2024-05-25
Some improvements

lbenet:


lbenet:
I think this is ready; in my opinion, the only major proposal in this PR is add another filter to the preliminary orbits obtained by Gauss.

LuEdRaMo:
Thanks @lbenet for this PR, as we discussed offline, we will save bumping the patch version to the upcoming PRs. Merging...

PR #69, opened 2024-06-21
Admissible Region Inner Boundary

LuEdRaMo:
This PR adds the possibility to plot an admissible region's inner boundary given by the constraint of non-negative geocentric energy.


LuEdRaMo:
Tests pass and this is ready to merge...

PR #70, opened 2024-06-23
Modified Target Plane

LuEdRaMo:
This PR adds the function `mtp` to compute the cartesian coordinates on the modified target plane of an asteroid's close approach with the Earth, according to [Farnocchia, et al. (2019)](https://doi.org/10.1007/s10569-019-9914-4).


LuEdRaMo:
Since this is a very straightforward PR and tests pass, I'll proceed to merge...

PR #71, opened 2024-07-03
WIP: Reject outliers following during orbital fit iterations

PerezHz:
This PR adds `outlier_rejection_carpino03` which performs the outlier rejection algorithm from Carpino et al. (2003) during orbital fit iterations (currently proof-of-concept). This PR also changes the current implementation of Carpino et al. scheme, e.g. in this line:
https://github.com/PerezHz/NEOs.jl/blob/2e0c6846e3f71e4e6b2b9b056491f9f27638b8b8/src/orbitdetermination/least_squares.jl#L99
where I reckon the weights should be inverted? i.e.:
```
γ = [inv(w_α) zero(T); zero(T) inv(w_δ)] 
```
(or at least the notation description at start of page 251 in Carpino's paper was enough to confuse me 😅). With this approach, the K-means clustering-based outlier rejection and the Carpino et al. (2003) scheme give essentially the same results as currently done in tests for 2007 VV7.

### DIFF
--- a/src/orbitdetermination/admissibleregion.jl
+++ b/src/orbitdetermination/admissibleregion.jl
@@ -308,7 +308,7 @@ function _georangerates(coeffs::Vector{T}, ρ::S) where {T, S <: Real}
     if d > 0
         return [-sqrt(arG(coeffs, ρ)), sqrt(arG(coeffs, ρ))]
     elseif d == 0
-        return zero(T) * arG(coeffs, ρ)
+        return zero(T) * [arG(coeffs, ρ)]
     else
         return Vector{T}(undef, 0)
     end
--- a/src/orbitdetermination/least_squares.jl
+++ b/src/orbitdetermination/least_squares.jl
@@ -52,7 +52,7 @@ end
 @doc raw"""
     carpino_smoothing(n::T) where {T <: Real}
 
-Fudge term for rejection condition in [`outlier_rejection`](@ref).
+Fudge term for rejection condition in [`ls_outrej_carpino03`](@ref).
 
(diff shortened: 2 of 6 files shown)


LuEdRaMo:
Thanks @PerezHz for this contribution. I have tested this in 7 asteroids where outlier rejection is needed and the results are consistent with both JPL and NEODyS in most cases. There are asteroids where the K-means clustering-based outlier rejection gives a better result, but that may be because we currently cannot modify the parameters of Carpino et al. (2003) from `outlier_rejection_carpino03`.  Is this ready for review or are you planning to add more?

PerezHz:
> There are asteroids where the K-means clustering-based outlier rejection gives a better result, ...

I think we should indeed look at this closer; what we know is that Carpino et al. (2003) gives very good results in the ensemble of known NEAs. If we can improve on that at the ensemble level that can be certainly helpful. 

>  ... that may be because we currently cannot modify the parameters of Carpino et al. (2003) from outlier_rejection_carpino03

That is already a good suggestion, I can definitely add the tunable parameters as `kwargs` so that they're accessible from `outlier_rejection_carpino03`.

> Is this ready for review or are you planning to add more?

I would like to add more things here, e.g., currently only least-squares fitting via Newton's method is supported for this outlier rejection scheme. So I would like to add support for differential corrections and Levenberg-Marquardt in a generic way, but that probably requires a bit of a redesign of the least-squares fitting code. That said, if `outlier_rejection_carpino03` is considered useful in its present form, we can try to merge it and do the other improvements afterwards. In any case I think we should give priority to #72.

PR #72, opened 2024-07-20
Update to TaylorSeries.jl v0.18

PerezHz:
See also PerezHz/TaylorIntegration.jl#194


PerezHz:
Tests are passing but there are a few things here to solve first:

- This PR depends on JuliaDiff/TaylorSeries.jl#364, so that will have to be merged first.
- This PR also depends on PerezHz/TaylorIntegration.jl#194. There seems to be an CI issue there, so that has to be fixed first as well before trying to merge this PR.
- This PR is currently of top of #71, will put it instead on top of `main` to decouple them

Project.toml
lbenet, line 1: Two comments:
- I think we do require the dependency on `TaylorIntegration` (due to `solcoeff!`, at least), i.e., VERSION >= v0.15.3
- `PlanetaryEphemeris` might need to also be upgraded

PerezHz: I fully agree, I just temporarily removed those dependencies to test some workarounds but they're back again.

PerezHz: And also did the update on PlanetaryEphemeris to work with TaylorSeries v0.18.

PerezHz: Fixed the compat entries as well 👍

PerezHz:
Tests are passing modulo coverage which diminished due to jetcoeffs! methods shrinking. I'll go ahead and merge.

PR #74, opened 2024-07-25
Add `neocp.jl`

LuEdRaMo:
This PR introduces a few functions to download data from the [Minor Planet Center NEO Confirmation Page](https://www.minorplanetcenter.net/iau/NEO/toconfirm_tabular.html).


test/observations.jl
```diff
@@ -218,6 +218,37 @@ using NEOs: src_path
...
+        using NEOs: NEOCPObject
+
+        # Fetch current NEOCP list
+        objects = fetch_objects_neocp()
+        @test isa(objects, Vector{NEOCPObject{Float64}})
+        # Pick the first object
+        object = objects[1]
+        # Check NEOCPObject fields
+        @test 0 <= object.score <= 100
+        @test object.date >= Date(2024)
```
lbenet, line 231:
```suggestion
        @test object.date >= year(today())
```

LuEdRaMo: I think something even better would be:
```julia
@test object.date >= today() - Month(3)
```
since we do not expect objects to be so long on the NEOCP.

lbenet: I fully agree; my point was simply to make something automatic, i.e., do not depend on "2024".

LuEdRaMo:
This is a very straightforward PR and tests pass, so I'll proceed to merge. I do not think this is worth tagging a new version.

PR #81, opened 2024-08-19
Improve time scale conversion-related code

PerezHz:
This PR improves docstrings and modifies internal code to make more explicit the internal conventions used to call `propagate` and time scale conversion related methods. Namely, the current convention is that `jd0`, the second argument in the call to `propagate`, is a Julian date in TDB time scale. This PR also deprecates `datetime2et` and renames it as `dtutc2et`, to make more explicit the fact that such method is meant to be used to transform a UTC time and date instantiated as a `DateTime`, and convert it to ET seconds, i.e., TDB seconds since J2000.0. This PR also adds a new method, `dtutc2jdtdb`, which transforms a UTC date and time (instantiated as a `DateTime`) to a TDB Julian date (i.e., JDTDB).

Potentially, some code in `AdmisibleRegion` and related methods might need to be updated, where `Dates.datetime2julian` is sometimes used to transform from a UTC date and time format to a Julian date in TDB, but where the new `NEOs.dtutc2jdtdb` method should be used instead.

Since the changes contained here are non-breaking, I propose to merge this (modulo any suggestions and improvements that can be always be added) and then tackle potential consistency issues related to internal use of `Dates.datetime2julian` in another PR.

cc: @lbenet @LuEdRaMo

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NEOs"
 uuid = "b41c07a2-2abb-11e9-070a-c3c1b239e7df"
 authors = ["Jorge A. Pérez Hernández", "Luis Benet", "Luis Eduardo Ramírez Montoya"]
-version = "0.8.11"
+version = "0.8.12"
 
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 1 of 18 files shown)


LuEdRaMo:
Thanks @PerezHz for solving this, I agree that we can merge this and update de OD interface in a future PR. Modulo rebasing to current main, it LGTM.

PR #91, opened 2024-12-05
New `research` folder

LuEdRaMo:
As suggested by @lbenet, this PR creates a new `research` folder to contain all scripts used in published papers.

PR #93, opened 2025-03-01
Fix numerical stability in `arG`

PerezHz:
This is just a proposal for a small fix on top of #91 to aid numerical stability of `arG`, which caused unit tests to fail locally on my machine (macOS), and also reproduced it in the CI by re-enabling macOS tests. An even better fix could be to add a numerical threshold in `arG` to check is we are close to the root of G which marks the boundary of the AR in the range, but we can look at that after merging #91. I have also (temporarily) re-enabled testing macOS. cc: @lbenet @LuEdRaMo

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,6 +23,7 @@ jobs:
           - 'nightly'
         os:
           - ubuntu-latest
+          - macos-latest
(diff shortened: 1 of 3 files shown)


LuEdRaMo:
Thanks @PerezHz for fixing this!

PR #94, opened 2025-03-08
Small fix in least squares code

PerezHz:
This is just a small fix  to dispatch on least squares method types using Julia's type system without resorting to `Symbol` for this. This also entailed reorganizing a bit some `include`s to make things work.


LuEdRaMo:
Thanks @PerezHz for this addition!

PR #102, opened 2025-06-04
Rename of subdirs in research

lbenet:
This is a rather small change to make life easier to browse the research directory. I do't think it is worth bumping a patch version (until the next one).


LuEdRaMo:
Looks good to me! Merging...

Issue #104, opened 2025-07-02
Add Marsden cometary model


PerezHz:
Currently NEOs.jl supports a model of non-gravitational accelerations based on `A1` and `A2` which represent resp. SRP and the Yarkovsky effect. For comets, a more general model is available:

https://articles.adsabs.harvard.edu/pdf/1973AJ.....78..211M

We should consider implementing this in NEOs.jl

PerezHz:
This might also point to figuring out a way to make our dynamical models more flexible without having to rewrite a full dynamical model (plus the overhead of setting up manually the associated jetcoeffs, etc.)

PR #108, opened 2025-08-03
Generalize `OpticalResidual` to include correlations

LuEdRaMo:
The main goal of this PR is to generalize the `OpticalResidual` struct to include correlations between the right ascension and declination residuals. In addition, this PR:

- Redefines the construction of optical tracklets when using the ADES format; the observations are now grouped by the `trkid` field instead of the observation night and station. The former enables the construction of nontrivial tracklets of satellite observations.
- Implements the Marsden et al. (1973) non gravitational accelerations model, as suggested in https://github.com/PerezHz/NEOs.jl/issues/104.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NEOs"
 uuid = "b41c07a2-2abb-11e9-070a-c3c1b239e7df"
 authors = ["Jorge A. Pérez Hernández", "Luis Benet", "Luis Eduardo Ramírez Montoya"]
-version = "0.15.0"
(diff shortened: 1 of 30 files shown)


LuEdRaMo:
Thanks for the suggestion! Those names are definitely more concise and overall better. In https://github.com/PerezHz/NEOs.jl/pull/108/commits/76bb4472a5dbcfa445c74440d4581f10f6609a76 I've renamed both dynamical models and added more informative docstrings. Having done that, I think this PR is ready to be merged...

PR #175, opened 2026-04-04
Improvements in memory in `t2c_jpl_de430` et al

lbenet:
See https://github.com/PerezHz/PlanetaryEphemeris.jl/pull/56


lbenet:
Pushing changes after rebasing... hopefully it solves the errors in the tests

LuEdRaMo:
Thanks for this contribution @lbenet!

> Pushing changes after rebasing... hopefully it solves the errors in the tests

The tests fail because Github installs the latest compatible version of PlanetaryEphemeris which does not yet contain the in-place method of `t2c_jpl_de430`. So, we can either wait until https://github.com/PerezHz/PlanetaryEphemeris.jl/pull/56 is merged and restart the tests here or you can modify the CI.yml to install the corresponding PE branch.

lbenet:
I think it's better to wait for https://github.com/PerezHz/PlanetaryEphemeris.jl/pull/56 to be merged.

src/propagation/buffer.jl
```diff
@@ -139,10 +143,13 @@ function PropagationBuffer(
         UJ_interaction = falses(N)
         # Turn on Earth interaction
         UJ_interaction[ea] = true
+        _orientation_ = allocate_c2t_jpl_de430(_acceph_.t)
     end
+    M_ = Array{Taylor1{U}}(undef, 3, 3, N)
+    zero_q_1 = auxzero(Taylor1([q0[1]], order))
```
LuEdRaMo: Given that `t2c_jpl_de430!` will update `M_` in-place, you need to initialize the corresponding entries of `M_`:
```Julia
for j in axes(M_, 2)
    for i in axes(M_, 1)
        M_[i, j, ea] = zero(zero_q_1)
    end
end
```

lbenet: This is (already) done, in a forthcomming commit (that I haven't pushed yet).

LuEdRaMo:
@lbenet Once the latest version of PlanetaryEphemeris (v0.11.4) gets registered, can you rebase this branch and bump the patch version?

lbenet:
> @lbenet Once the latest version of PlanetaryEphemeris (v0.11.4) gets registered, can you rebase this branch and bump the patch version?

Sure. I guess also tagging a patch version is enough, right?

LuEdRaMo:
> Sure. I guess also tagging a patch version is enough, right?

Yes, but I will tag the new version after merging...

If you consider this is ready, I will merge once the tests pass.

lbenet:
> If you consider this is ready, I will merge once the tests pass.

I think this is ready, but I oversaw a small error that needs to be fixed in PlanetaryEphemeris.jl; I'll correct it in a new PR. I guess a new patch version will be needed...

LuEdRaMo:
@lbenet Can you update the PlanetaryEphemeris compat entry? 

Actually, I think it would be better to set it to 0.11, not specifying the patch version. The same with TaylorSeries.

lbenet:
Sure, I'll do it. I actually specified the patch versions because previous versions wouldn't work. I push the commit shortly.

LuEdRaMo:
The values in two propagation tests decreased by a few thousandths, so I've updated the corresponding tests. I think this is expected given the small change we noted in the computation of the rotation matrix from PlanetaryEphemeris.

LuEdRaMo:
@lbenet Do you have any further comments on this PR, before merging?

lbenet:
No further comments, I think this is ready, but there is one test still broken. Any ideas?

lbenet:
Seems a questions of a rejected connection...

LuEdRaMo:
> No further comments, I think this is ready, but there is one test still broken. Any ideas?

As part of the tests we download observations from NEODyS-2, and it seems that the NEODyS-2 server is down, or at least I cannot connect. We just have to wait until it resolves and then rerun the tests.

LuEdRaMo:
NEODyS-2 server is up again and tests pass.... Merging...

P.D. Running the impact monitoring benchmarks it looks like this PR reduces allocations by ~50%.

lbenet:
Thanks a lot!

PR #206, opened 2026-05-31
Update to PlanetaryEphemeris 0.12

PerezHz:
This PR also bumps TaylorSeries compat bounds to v0.22


LuEdRaMo (commented):
Thanks @PerezHz for looking into this! 

For now I only have one comment, the scripts in the `research` folder are tied to a particular version of NEOs (and therefore a particular version of TS and PE) via their projects. Thus, we should not update the TS API for any script whose NEOs compat entry is prior to the upcoming 0.27...

PerezHz:
@LuEdRaMo thanks for the input, so should I leave research scripts untouched? Are there any other scripts it would be better to not touch at this time?

PerezHz:
@LuEdRaMo @lbenet this is now ready for you to have a look; this update mostly went fine, and thanks a lot @LuEdRaMo for running the update for the planetary ephemerides! I have uploaded the new ephemerides file, see [here](https://github.com/PerezHz/sseph). I did some minor changes on TI which have been released and the last few CI runs here are already using those changes, as well as PE v0.12. I was also updating the scripts in `research/` and `scripts/` to use the new TaylorSeries `variables` and `variables!` but undid those changes following @LuEdRaMo's comment on the matter.

PerezHz:
I pushed a commit where I'm testing a new guardrail in `timeindex` for `TaylorSolution` which covers an edge case which under normal circumstances should not happen since this is now covered for at `TaylorSolution` construction, but it's better anyway to throw an error there just in case. If testing goes fine, will release a patch version for TI and will update this PR to not use a custom TI branch.

PerezHz:
Tests passed on TI custom branch, released TI v0.18.10, CI now re-running, so this is still ready for review (unless CI tests don't pass, but they should)

UPDATE: tests pass! (except coverage)

LuEdRaMo:
Thanks again @PerezHz for this update! I hope to review this tomorrow. 

I've merged https://github.com/PerezHz/NEOs.jl/pull/202, which only included a few minor bug fixes and updates. Can you rebase this please?

test/astrometry.jl
```diff
@@ -266,7 +266,6 @@ using NEOs: SCRATCH_PATH
         @test obs["rhocosphi"] == "0.845107"
         @test obs["rhosinphi"] == "0.533611"
         @test obs["short_name"] == "University of Arizona Mt. Lemmon Survey"
-        @test obs["updated_at"] == "Tue, 26 May 2026 20:34:56 GMT"
```
PerezHz, line 269: I removed this test because this can change rather frequently and it's not crucial to test for the exact value. I think we could consider instead testing that it is a string and/or is a date with some specific format.

PerezHz:
@LuEdRaMo I updated to current `main`

notebooks/Apophis.ipynb
```diff
@@ -49,7 +49,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dq = set_variables(\"δx\", order=5, numvars=7)"
+    "dq = TaylorSeries.variables!(\"δx\", order=5, numvars=7)"
```
LuEdRaMo: The files in the `notebooks` folder have not been updated since the early versions of NEOs; in fact, this notebook uses functions that no longer exist. So, we can either update these notebooks or, analogously to the scripts in `research`, link them to a particular version of the package.

notebooks/B-plane-jt-o5v7.ipynb
```diff
@@ -74,7 +74,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dq = set_variables(\"ξ\", order=5, numvars=7)"
+    "dq = TaylorSeries.variables!(\"ξ\", order=5, numvars=7)"
```
LuEdRaMo: Same comment as above.

pha/bennu.jl
```diff
@@ -4,6 +4,7 @@
 #julia --project=@. main.jl
 using NEOs
 using Dates
+using TaylorSeries
```
LuEdRaMo: Again, this script has been practically unchanged since the early versions of NEOs and it would not work with the current version because of the signature of `propagate`. @PerezHz, since you wrote these scripts (and notebooks) I think you should decide whether we update them or leave them untouched.

scripts/findorbit/findorbit.jl
```diff
@@ -26,16 +27,52 @@ function parse_commandline()
...
+    fmt = normalize_astrometry_format(format)
+    if isfile(input)
+        fmt = fmt == "auto" ? detect_astrometry_format(input) : fmt
+        optical = fmt == "ades" ? read_optical_ades(input) : read_optical_mpc80(input)
+    else
+        fmt = fmt == "auto" ? "ades" : fmt
+        optical = fmt == "ades" ? fetch_optical_ades(input, MPC) : fetch_optical_mpc80(input, MPC)
+    end
+    return optical, fmt
+end
```
LuEdRaMo: I suggest we define a function in NEOs called `read_optical_astrometry` that does this,  i.e., read an arbitrary file and deduce the observations' format. The reason that has not been done so far is because it is an intrinsically type-unstable operation; yet, given that reading the astrometry is far from the most expensive part of any process, I think we can absorb the cost of the type-instability. Although that may be the subject of another PR...

scripts/findorbit/findorbit.jl
```diff
@@ -222,25 +261,35 @@ function main()
...
         # Multiple apparition orbit determination
-        orbitMA = multipleapparition(apps, orbitMA, params)
+        orbit = multipleapparition(apps, orbit, params)
     end
 
     # Shift epoch to the middle of the observational arc
-    tmean = meanepoch(orbitMA)
-    orbitMA = shiftepoch(orbitMA, tmean + PE.J2000, params)
+    tmean = meanepoch(orbit)
+    orbit = shiftepoch(orbit, tmean + PE.J2000, params)
```
LuEdRaMo: I understand that these changes make the code more readable, but they introduce a type-instability as the result of `single apparition` is a `SingleApparitionOrbit` but `bridge` (and subsequent functions) return a `MultipleApparitionOrbit`. We should check if such type-instability affects or not the performance of the script.

scripts/main.jl
```diff
@@ -4,6 +4,7 @@
 #julia --project=@. main.jl
 using NEOs
 using Dates
+using TaylorSeries
```
LuEdRaMo: This is another script that has long been unmaintained, so the same applies as in the previous  comments...

src/impactmonitoring/closeapproach.jl
```diff
@@ -57,13 +57,18 @@ nominalstate(x::CloseApproach) = [cte(x.x), cte(x.y), cte(x.z)]
 
 difft(x::CloseApproach, y::CloseApproach) = abs(nominaltime(x) - nominaltime(y))
 
-get_order(x::CloseApproach{T, U}) where {T, U <: AbstractSeries} = get_order(x.t)
+taylor_order(x::CloseApproach{T, U}) where {T, U <: AbstractSeries} = taylor_order(x.t)
+if isdefined(TaylorSeries, :order)
+    TaylorSeries.order(x::CloseApproach{T, U}) where {T, U <: AbstractSeries} = taylor_order(x)
+else
+    TaylorSeries.get_order(x::CloseApproach{T, U}) where {T, U <: AbstractSeries} = taylor_order(x)
+end
```
LuEdRaMo: What is the difference between `taylor_order` and `order`/`get_order`? AFAIK the former is meant to replace the latter. Given that this is going to be a breaking release that requires the new versions of TS, TI and PE, are we keeping the latter for backwards compatibility?

src/impactmonitoring/lov.jl
```diff
@@ -53,7 +58,7 @@ function LineOfVariationsBuffer(IM::AbstractIMProblem{D, T}, lovorder::Int,
     end
     # Initial condition
     q00 = orbit()
-    q0TN = q00 + sigmas(orbit) .* get_variables(T, lovorder)
+    q0TN = q00 + sigmas(orbit) .* taylor_variables(T, lovorder)
```
LuEdRaMo: If I understand correctly, this line is using the global TS variables. If that is the case, this PR is sort of an intermediate step towards updating NEOs to the new version of TS that no longer supports global jet spaces. Am I correct?

src/impactmonitoring/lov.jl
```diff
@@ -150,10 +160,10 @@ function covariance(
     mjd0 = t0 + MJD2000
     jd0 = t0 + PE.J2000
     # Order with respect to LOV index
-    order = get_order(buffer)
+    order = taylor_order(buffer)
     # Jet transpot initial condition
     car00 = lovtransform(mjd0, coord00, sun, Val(coord), Val(:cartesian))
-    carTN = car00 + scalings .* get_variables(T, order)
+    carTN = car00 + scalings .* taylor_variables(T, order)
```
LuEdRaMo: Same comment as above.

src/impactmonitoring/virtualasteroid.jl
```diff
@@ -19,7 +19,7 @@ end
 
 numtypes(::VirtualAsteroid{T, U}) where {T, U} = T, U
 
-# This function assumes that order ≤ get_order(x)
+# This function assumes that order ≤ TaylorSeries.order(x)
```
LuEdRaMo, line 22: Shouldn't this be `TaylorSeries.taylor_order(x)`?

src/impactmonitoring/virtualasteroid.jl
```diff
@@ -46,15 +46,20 @@ nominaltime(x::VirtualAsteroid) = x.epoch
...
+    TaylorSeries.get_order(x::VirtualAsteroid{T, U}) where {T, U <: AbstractSeries} = taylor_order(x)
+end
 
 """
     virtualasteroids(lov, method [, vaorder]; kwargs...)
 
 Divide the line of variations `lov` into a vector of `Taylor1` virtual
 asteroids using a given sampling `method` and truncating the polynomials
-to degree `vaorder` (default: `get_order(lov)`). Each of the following
+to degree `vaorder` (default: `TaylorSeries.order(lov)`). Each of the following
```
LuEdRaMo, line 57: Same comment as above.

src/orbitdetermination/abstractorbit/abstractorbit.jl
```diff
@@ -31,7 +31,7 @@ numvars(x::AbstractOrbit) = length(x.variables)
 
 Return the reference epoch of an orbit in TDB days since J2000.
 """
-epoch(x::AbstractOrbit) = x.bwd.t0
+epoch(x::AbstractOrbit) = first(x.bwd.t)
```
LuEdRaMo: I notice we do this operation very often throughout NEOs. Maybe we should define (on TI) something like `firsttime(::TaylorSolution)` and `lasttime(::TaylorSolution)` and then define methods of such functions for all NEOs' objects that contain `TaylorSolution`s. Again, we can leave that to another PR...

src/orbitdetermination/abstractorbit/leastsquaresorbit.jl
```diff
@@ -112,21 +112,21 @@ An asteroid least squares orbit.
...
             dynamics::D, variables::Vector{Int}, optical::O, tracklets::TrackletVector{T},
-            radar::R, bwd::TaylorInterpolant{T, U, 2}, fwd::TaylorInterpolant{T, U, 2},
+            radar::R, bwd::DensePropagation2{T, U}, fwd::DensePropagation2{T, U},
             ores::Vector{OpticalResidual{T, U}}, rres::RR, fit::LeastSquaresFit{T},
             qs::Matrix{T}, Qs::Vector{T}
         ) where {
             D, T <: Real, U <: Number, O <: AbstractOpticalVector{T},
             R <: Union{Nothing, AbstractRadarVector{T}},
             RR <: Union{Nothing, Vector{RadarResidual{T, U}}}
         }
```
LuEdRaMo, line 122: In principle , I think we can relax the type constraints and construct orbits from `TaylorSolution`s with views, which might be helpful in the future. However, that will be a very delicate operation that we can tackle in future PRs.

src/orbitdetermination/abstractorbit/leastsquaresorbit.jl
```diff
@@ -243,18 +243,18 @@ function zero(::Type{LeastSquaresOrbit{D, T, U, O, R, RR}}) where {D, T, U, O, R
                              ores, rres, fit, qs, Qs)
 end
 
-iszero(x::LeastSquaresOrbit{D, T, U, O, R, RR}) where {D, T, U, O, R, RR} =
-    x == zero(LeastSquaresOrbit{D, T, U, O, R, RR})
+iszero(x::LeastSquaresOrbit) = isempty(x.variables) && isempty(x.ores) &&
+    iszero(x.bwd) && iszero(x.fwd)
```
LuEdRaMo: I would suggest:
```julia
iszero(x::LeastSquaresOrbit) = x == zero(typeof(x))
```
Or, has something changed in the equality comparisons of `TaylorSolution`s that will make the above not work?

src/orbitdetermination/abstractorbit/preliminaryorbit.jl
```diff
@@ -79,17 +79,17 @@ A preliminary orbit computed by Gauss method.
     # Inner constructor
     function GaussOrbit{D, T, U, O}(
             dynamics::D, variables::Vector{Int}, optical::O, tracklets::TrackletVector{T},
-            bwd::TaylorInterpolant{T, U, 2}, fwd::TaylorInterpolant{T, U, 2},
+            bwd::DensePropagation2{T, U}, fwd::DensePropagation2{T, U},
```
LuEdRaMo, line 82: Same comments apply for this file as in `leastsquaresorbit.jl`

src/orbitdetermination/odproblem.jl
```diff
@@ -137,8 +137,8 @@ function init_radar_residuals(::Type{U}, od::MixedODProblem, idxs = radarindices
 end
 
 function set_od_order(::Type{T}, varorder::Int, numvars::Int = 6) where {T <: Real}
-    if get_order() < varorder || get_numvars() != numvars
-        set_variables(T, "dx"; order = varorder, numvars = numvars)
+    if taylor_order() < varorder || get_numvars() != numvars
+        taylor_variables!(T, "dx"; order = varorder, numvars = numvars)
     end
     return nothing
 end
```
LuEdRaMo, line 144: This function was introduced to avoid resetting the global variables if the current jet space was compatible with the new (had more variables and a greater than or equal order). If the recent changes in TS have significantly reduced the latency of updating the global jet space, we might as well deprecate this function in favour of `taylor_variables!`.

src/propagation/propagation.jl
```diff
@@ -37,14 +37,14 @@ end
 
 Equivalent to:
 
-`TaylorSeries.set_variables(T, names; order, numvars = length(c))`
+`TaylorSeries.variables!(T, names; order, numvars = length(c))`
```
LuEdRaMo, line 40: Similarly to a previous comment, what is the difference between `variables!` and `taylor_variables`?

src/NEOs.jl
```diff
@@ -45,9 +45,21 @@ using SpecialFunctions: erf
...
+const _taylor_variables =
+    getfield(TaylorSeries, isdefined(TaylorSeries, :variables) ? :variables : :get_variables)
+const _taylor_variables_bang =
+    getfield(TaylorSeries, isdefined(TaylorSeries, Symbol("variables!")) ?
+        Symbol("variables!") : :set_variables)
+
+taylor_order(args...; kwargs...) = _taylor_order(args...; kwargs...)
+taylor_variables(args...; kwargs...) = _taylor_variables(args...; kwargs...)
+taylor_variables!(args...; kwargs...) = _taylor_variables_bang(args...; kwargs...)
+
```
LuEdRaMo: This block answers many of my questions above. Still, the question is, should we worry about backwards compatibility given that this version of NEOs will not work with previous versions of TS, TI and PE?

src/parameters.jl
```diff
@@ -91,8 +91,8 @@ nongravitational accelerations model:
     marsden_scalings::NTuple{3, T} = (0.0, 0.0, 0.0)
     marsden_radial::NTuple{5, T} = (1.0, 1.0, 2.0, 0.0, 0.0)
     # Sun (earth) ephemeris
-    eph_su::TaylorInterpolant{T, T, 2, Vector{T}, Matrix{Taylor1{T}}} = _loadeph(su)
-    eph_ea::TaylorInterpolant{T, T, 2, Vector{T}, Matrix{Taylor1{T}}} = _loadeph(ea)
+    eph_su::DensePropagation2{T, T} = _loadeph(su)
+    eph_ea::DensePropagation2{T, T} = _loadeph(ea)
```
LuEdRaMo: Given how simple `_loadeph` became, I would simply write:
```Julia
eph_su::DensePropagation2{T, T} = selecteph(sseph, su)
eph_ea::DensePropagation2{T, T} = selecteph(sseph, ea)
```

test/astrometry.jl
LuEdRaMo: These tests are ment to evaluate whether we communicate correctly with the MPC's observatories API. I would suggest to either comment out this particular test or do something like
```Julia
@test obs["updated_at"] >= "Tue, 26 May 2026 20:34:56 GMT"
```

test/propagation.jl
```diff
@@ -325,32 +322,30 @@ end
...
-        @test mean(res) ≈ -0.0140 atol=1e-4
-        @test std(res) ≈ 0.3655 atol=1e-4
-        @test chi2(res) ≈ 121.6226 atol=1e-4
-        @test nms(res) ≈ 0.1337 atol=1e-4
-        @test nrms(res) ≈ 0.3656 atol=1e-4
+        @test mean(res) ≈ -0.0139 atol=1e-4
+        @test std(res) ≈ 0.3650 atol=1e-4
+        @test chi2(res) ≈ 121.3048 atol=1e-4
+        @test nms(res) ≈ 0.1333 atol=1e-4
+        @test nrms(res) ≈ 0.3651 atol=1e-4
```
LuEdRaMo, line 343: My guess is that these changes are mainly a consequence of updating the Solar System ephemerides. Still, it looks like they are within the expected bounds.

test/propagation.jl
```diff
@@ -373,23 +368,17 @@ end
...
         # TODO: fix roundoff differences near deep close approach in 2029
-        @test norm(sol.x - solnp.x, Inf) / norm(solnp.x, Inf) < 4e-18 # 3.757708512785821e-20
-        @test JLD2.writeas(typeof(sol)) == TaylorInterpolantNSerialization{Float64}
-        jldsave("test.jld2"; sol)
-        recovered_sol = JLD2.load("test.jld2", "sol")
-        @test sol == recovered_sol
-        rm("test.jld2")
+        @test norm(sol.p - solnp.p, Inf) / norm(solnp.p, Inf) < 4e-18 # 3.757708512785821e-20
+        # TaylorN-valued TaylorSolution serialization currently needs an upstream
+        # JLD2/TaylorSeries fix; scalar TaylorN serialization is covered below.
```
LuEdRaMo: Why can we no longer to this comparison and the ones below? 

A related question, if `TaylorSolution` has fields of type `Nothing`, how does TI handle the corresponding equality comparison? To decide if two `TaylorSolution`s are equal, shouldn't we compare only the non `nothing` fields?

LuEdRaMo:
@PerezHz I left a couple of comments (which are mostly questions). Apart from that, this update looks great!

test/astrometry.jl
PerezHz: Ok, I'll leave this test commented for now; if the MPC obscodes API is somehow unreachable this will be visible from the previous tests.

notebooks/Apophis.ipynb
PerezHz: Thanks for spotting this. I undid the changes here; I'd suggest tackling updating the notebooks in another PR.

notebooks/B-plane-jt-o5v7.ipynb
PerezHz: Ok, changes undone, let's leave this for another PR.

pha/bennu.jl
PerezHz: Changes undone so that we instead update them in another PR.

src/NEOs.jl
PerezHz: Thank you for the comment. I don't think we should worry about backwards compatibility; these were meant as an intermediate step for compatibility before releasing TS 0.22. I've removed them now, so code now uses `TS.order`, `TS.variables`, `TS.variables!`.

src/parameters.jl
PerezHz: Thanks, done

src/orbitdetermination/odproblem.jl
PerezHz: See https://github.com/PerezHz/NEOs.jl/pull/206#discussion_r3350729751

src/propagation/propagation.jl
PerezHz: See https://github.com/PerezHz/NEOs.jl/pull/206#discussion_r3350729751

scripts/main.jl
PerezHz: Changes here undone

src/impactmonitoring/closeapproach.jl
PerezHz: `taylor_order`, etc., were meant as intermediate steps before release of TS 0.22; I removed them. See https://github.com/PerezHz/NEOs.jl/pull/206#discussion_r3350729751

src/impactmonitoring/lov.jl
PerezHz: Switched to `TS.variables`

src/impactmonitoring/lov.jl
PerezHz: Switched to `TS.variables`

src/impactmonitoring/virtualasteroid.jl
PerezHz: Removed `taylor_order`; see https://github.com/PerezHz/NEOs.jl/pull/206#discussion_r3350729751

src/impactmonitoring/virtualasteroid.jl
PerezHz: Fixed; see https://github.com/PerezHz/NEOs.jl/pull/206#discussion_r3350729751

test/propagation.jl
PerezHz: Indeed, the variations are within the expected bounds, so only a small tweaking was needed here. this is an effect of switching from relative to absolute times in the dense propagation structure (i.e,., from `TaylorInterpolant` from `TaylorSolution`). These tests are really sensitive, and to me they actually show the robustness of the AD approach to compute radar observables (Doppler).

test/propagation.jl
PerezHz: Thanks for spotting this; at some point I was testing `TaylorSolution` and forgot to bring these back, but they're back now (and passing)

PerezHz: About `TaylorSolution` equality: it should work with `Nothing` fields. `TaylorIntegration` defines `TaylorSolution` with `@auto_hash_equals`, so equality is full-field equality. `nothing == nothing` is `true,` and `nothing == some_array` is `false`. I think that is the correct behavior: a dense solution with `p` should not compare equal to a non-dense solution with `p = nothing`.

src/orbitdetermination/abstractorbit/leastsquaresorbit.jl
PerezHz: Thanks for the suggestion, I added it and it works!

src/orbitdetermination/abstractorbit/leastsquaresorbit.jl
PerezHz: Yes, I agree this would be useful, and makes sense to tackle in another PR

src/orbitdetermination/abstractorbit/preliminaryorbit.jl
PerezHz: Fixed `iszero` here as well

src/orbitdetermination/abstractorbit/abstractorbit.jl
PerezHz: Thanks for this suggestion! I added this in `TaylorIntegration` and pushed the update here to use `firsttime`, `lastime` which I agree are really helpful

scripts/findorbit/findorbit.jl
PerezHz: Thanks for this suggestion; I added a function barrier which should help with potential type instabilities here. Since the classification per apparition type (single, bridge, multiple) is done once per script run, this should not impact performance by a lot anyway (or at least that's my current estimate).

scripts/findorbit/findorbit.jl
PerezHz: Thanks for this suggestion, I have attempted to add it in my last commit and I think it works fine, though maybe it can be improved somehow

Project.toml
```diff
@@ -82,7 +82,7 @@ SpecialFunctions = "2"
 StaticArraysCore = "1.4"
 StatsBase = "0.33, 0.34"
 Tables = "1.10"
-TaylorIntegration = "0.18"
+TaylorIntegration = "0.18.11"
```
LuEdRaMo, line 85: I understand that this particular patch version is needed to import `firsttime` and `pasttime` from TI, is that correct?

LuEdRaMo:
@PerezHz I see that you have addressed all my reviews, thank you for that! 

I only have one final question regarding the TI compat entry; apart from that, if you think this is ready feel free to merge.

Project.toml
PerezHz: Exactly, yes

PerezHz:
@LuEdRaMo thanks for reviewing! I just wanted to mention something about the type instabilities: since those depend on runtime inputs from the user I think we cannot get rid of them completely, but at least the function barriers contain that instability to a single place. At the same time, they do not happen inside heavy numerical loops, but before the actual processing, so those runtime dispatches should not affect performance a lot.

So I'll go ahead and merge, thanks a lot!

PR #211, opened 2026-06-16
Documentation build via GitHub Actions

LuEdRaMo:
@PerezHz This PR is the final step needed to deploy the documentation with GitHub Actions (see https://github.com/PerezHz/NEOs.jl/issues/28). Yet, since I do not have access to this repository's settings, I need you to check a few things. First, we need the corresponding keys, as specified [here](https://m3g.github.io/JuliaNotes.jl/stable/publish_docs/#Add-the-keys-to-the-github-repository). Second, we must configure a publishing source for GitHub Actions, as explained [here](https://documenter.juliadocs.org/stable/man/hosting/#gh-pages-Branch). If you notice anything else this PR might be missing, let me know...

PR #212, opened 2026-06-29
Add fallback server to download EOP

LuEdRaMo:


PerezHz:
@LuEdRaMo thanks a lot for fixing this, looks great! I just now thinking that we should've gotten this one merged before #211 😅 (docs build/CI is failing on main due to the very issue this PR fixes)

PerezHz:
Also it looks like some numerical thresholds for radar astrometry tests are sensitive enough to changes in the EOP files, I'd suggest to fix those by increasing tolerances by an order of magnitude, so that they can include computed values with either set of EOP files.

LuEdRaMo:
> Also it looks like some numerical thresholds for radar astrometry tests are sensitive enough to changes in the EOP files, I'd suggest to fix those by increasing tolerances by an order of magnitude, so that they can include computed values with either set of EOP files.

Yeah, I was about to suggest that. If you want, feel free to update the tolerances in this PR and rebase to main. Otherwise, I will take care of it in a bit...

PR #219, opened 2026-07-04
Fix ADES parsing on XML.jl 0.4 and widen XML compat to 0.3, 0.4

mathieu17g:
[XML.jl v0.4](https://github.com/JuliaData/XML.jl/releases/tag/v0.4.0) (a breaking release, now registered) preserves inter-element whitespace as `Text` nodes and removed `LazyNode` iteration. `parse_optical_ades` relies on both, in three places: `node[2]` (no longer the `<ades>` element — it becomes a whitespace `Text` node, so parsing indented ADES silently yields **0 observations**), unfiltered `children` iteration (hits `Text` nodes), and `first(::LazyNode)` (no longer defined).

This PR locates the root and the observation entries by node type, filters per-observation children to elements, and reads leaf values through `children()` — all backwards-compatible with XML 0.3 — and widens `[compat]` to `XML = "0.3, 0.4"`.

> ⚠️ **Heads-up on #218** (CompatHelper): it widens the compat the same way but without the code fix — merging it alone would let the resolver pick XML 0.4 and silently break ADES parsing (0 observations, no error). This PR includes the compat change, so it can land instead of (or before) #218.

Verified on `test/data/433.xml` (17 020 observations):
- **XML 0.3.9** — the patched code extracts data identical to the current code (no behavior change).
- **XML 0.4.0** — the current code returns 0 observations; the patched code extracts the same 17 020-observation dataset as 0.3.9.

Happy to adjust anything.


LuEdRaMo:
Thanks a lot @mathieu17g for this contribution! I have no comments.

If you could just bump the patch version, I can merge this and tag the new version.

mathieu17g:
Done — bumped to 0.27.4. Thanks for the quick review!

LuEdRaMo:
Thanks again! Merging...

